PR #34, opened 2015-02-11
Remove portable target

bricelam:
Given the removal of ANIs, this isn't needed anymore


bricelam:
(This depends on aspnet/EntityFramework#1682)

Issue #56, opened 2015-04-14
Consider supporting arrays


gingters:
I could PR support for array types, if desired, based on what I did in my ConfigurationMapper: https://github.com/gingters/SmartDev.ConfigurationMapper/blob/dev/src/SmartDev.ConfigurationMapper/ValueTypeConverter.cs

Any interest in array support? If yes I'll prepare a PR.

davidfowl:
Configuration actually doesn't; support arrays currently. Discussion here https://github.com/aspnet/Configuration/issues/115

gingters:
Pretty much no input format except JSON does support arrays there.
The question would be more, would you treat specially formatted values (i.e. comma separated values) as an array and simply de-serialize these into an array?

victorhurdugaci:
Fixed in https://github.com/aspnet/Options/pull/59

Issue #55, opened 2015-04-10
Don't ignore errors in options when reading from config


Eilon:
Currently all errors are ignored: https://github.com/aspnet/Options/commit/e0ebe3cccf70e3adea3be421df774583ba6f047d (from https://github.com/aspnet/Options/issues/26)

This seems like it'll be difficult for users to debug.

If the config is invalid, the operation needs to fail.

victorhurdugaci:
This is done

PR #59, opened 2015-04-27
List and Dictionary binding

victorhurdugaci:
Fixes: https://github.com/aspnet/Options/issues/56

- Support for binding to lists
- Support for binding to dictionary

Please review: @lodejard @HaoK @glennc


victorhurdugaci:
Btw, Travis and AppVeyor will fail without my changes from Configuration. So don't panic :)

src/Microsoft.Framework.OptionsModel/ConfigurationBinder.cs
```diff
@@ -0,0 +1,290 @@
...
+                var hasPublicSetter = property.SetMethod != null && property.SetMethod.IsPublic;
+
+                if (propertyValue == null && !hasPublicSetter)
+                {
+                    // Property doesn't have a value and we cannot set it so there is no
+                    // point in going further down the graph
+                    return;
+                }
+
+                VisitType(property.PropertyType, ref propertyValue);
```
HaoK: nit: this is really VisitProperty right?

HaoK: Or more specifically, Read/GetPropertyValue?  Is there a reason this method is void with ref object instead of just returning the propertyValue?

victorhurdugaci: Ref is needed because you can have a get only property that was initialized previously and we want to be able to the bindings on that instance. So, it will use the existing instance if it exists, otherwise it creates a new one

victorhurdugaci: It could simply return the value but I think it is more explicit to have a ref parameter. It tells you that we can change that value

HaoK: Why can't you just return the reference to the previous instance or the new instance?

src/Microsoft.Framework.OptionsModel/ConfigurationBinder.cs
```diff
@@ -0,0 +1,290 @@
...
+
+        private IConfiguration _currentConfigurationContext;
+
+        public ConfigurationBinder([NotNull] IConfiguration configuration)
+        {
+            _context = new Stack<IConfiguration>();
+            _startConfiguration = configuration;
+        }
+
+        public static void Bind(IConfiguration configuration, object merge)
```
Eilon: I don't quite understand the names "Bind" and "Read". Are there better names for these? Also the "merge" parameter doesn't make sense to me...

src/Microsoft.Framework.OptionsModel/ConfigurationBinder.cs
HaoK: Yeah I think that is normally understood that objects are mutable.  I don't think we typically use ref parameters in general :)  It definitely sticks out a bit like a sore thumb... I'd just call it out in a comment for the method if you want to emphasize this point.  Making the signature void/ref is overkill (especially since this is a private method)...

src/Microsoft.Framework.OptionsModel/ConfigurationBinder.cs
```diff
@@ -0,0 +1,290 @@
...
+                        {
+                            // Something else
+                            VisitObject(typeInstance);
+                        }
+                    }
+                }
+            }
+        }
+
+        private void VisitIDictionaryInstance(object dictionary, Type iDictionaryType)
```
HaoK: Nit: VisitDictionary might be a better name (more concise)

src/Microsoft.Framework.OptionsModel/ConfigurationBinder.cs
```diff
@@ -0,0 +1,290 @@
...
+                {
+                }
+                finally
+                {
+                    ExitContext();
+                }
+            }
+        }
+
+        private void VisitICollectionTypeInstance(object collection, Type iCollectionType)
```
HaoK: nit: VisitCollection?

HaoK:
Should we consider moving the ConfigurationBinder to configuration itself?  For full disclosure, when I asked @lodejard this way back when I first wrote this, his reasoning was having the binding functionality live in options, rather than config nudges people towards options instead of directly binding against configuration...

src/Microsoft.Framework.OptionsModel/ConfigurationBinder.cs
HaoK: yeah, my guess is merge means the existing instance that gets the values from config merged into.  So perhaps something like targetInstance would be better

src/Microsoft.Framework.OptionsModel/ConfigurationBinder.cs
```diff
@@ -0,0 +1,290 @@
...
+            _context = new Stack<IConfiguration>();
+            _startConfiguration = configuration;
+        }
+
+        public static void Bind(IConfiguration configuration, object merge)
+        {
+            new ConfigurationBinder(configuration).Read(merge);
+        }
+
+        public void Read(object merge)
```
HaoK: Perhaps Bind(object instance) would be more clear?

src/Microsoft.Framework.OptionsModel/ConfigurationBinder.cs
```diff
@@ -0,0 +1,290 @@
...
+        }
+
+        public TOptions Read<TOptions>() where TOptions : new()
+        {
+            var options = new TOptions();
+            Read(options);
+            return options;
+        }
+
+        private void VisitObject(object obj)
```
HaoK: I don't think Visit really gives much meaning, would Bind be a better prefix for all of the Visit methods if that's what  each method is doing?  Something like BindObject, BindProperty, BindDictionary, etc...

test/Microsoft.Framework.OptionsModel.Test/ConfigurationCollectionBindingTests.cs
```diff
@@ -0,0 +1,299 @@
...
+            }
+        }
+
+        private class CustomDictionary<T>: Dictionary<string, T>
+        {
+        }
+
+        private class NestedOptions
+        {
+            public int Integer { get; set; }
```
HaoK: Can we also nest a List/Dictionary in here as well just to make sure that also works?

src/Microsoft.Framework.OptionsModel/ConfigurationBinder.cs
victorhurdugaci: Forgot to rename those reads. `Bind` should be everywhere because this code binds a configuration object to an options model.

src/Microsoft.Framework.OptionsModel/ConfigurationBinder.cs
victorhurdugaci: That's a good suggestion

victorhurdugaci:
Updated the PR

> Should we consider moving the ConfigurationBinder to configuration itself? For full disclosure, when I asked @lodejard this way back when I first wrote this, his reasoning was having the binding functionality live in options, rather than config nudges people towards options instead of directly binding against configuration...

mmm... maybe? I see the point of having bindings directly in configuration but at the same time having it in options makes sense too. Maybe it would make sense to merge options and configuration completely...

src/Microsoft.Framework.OptionsModel/ConfigurationBinder.cs
```diff
@@ -0,0 +1,284 @@
...
+        private readonly IConfiguration _startConfiguration;
+
+        private IConfiguration _currentConfigurationContext;
+
+        public ConfigurationBinder([NotNull] IConfiguration configuration)
+        {
+            _startConfiguration = configuration;
+        }
+
+        public static void Bind(IConfiguration configuration, object merge)
```
HaoK: merge => model?

src/Microsoft.Framework.OptionsModel/ConfigurationBinder.cs
```diff
@@ -0,0 +1,284 @@
...
+                }
+            }
+            finally
+            {
+                ExitContext();
+            }
+        }
+
+
+        private object BindType(Type type, object typeInstance)
```
HaoK: Can we give this a better name for exactly what this is doing?  Its Binding and merging an existing property? BindType just doesn't really say about what this is actually doing...

src/Microsoft.Framework.OptionsModel/ConfigurationBinder.cs
```diff
@@ -0,0 +1,284 @@
...
+                    return Convert.ChangeType(_currentConfigurationContext.Get(null), type);
+                }
+            }
+            catch
+            {
+                return null;
+            }
+        }
+
+        private void EnterContext(IConfiguration context)
```
HaoK: So looking at this code, do you really need a stack of IConfiguration?  Can all of the helpers be made static with the _currentConfigurationContext passed via a parameter instead?  I think it reads a bit easier overall without the Stack of IConfigurations unless they are actually needed for a specific reason...

src/Microsoft.Framework.OptionsModel/ConfigurationBinder.cs
victorhurdugaci: It has nothing to do with properties. It just binds the current configuration node (and subnodes) to a type. If you pass an instance of that type it will populate that, if not it tries to create a new instance. It is really `BindType` :)

src/Microsoft.Framework.OptionsModel/ConfigurationBinder.cs
victorhurdugaci: Sure

victorhurdugaci:
Updated

src/Microsoft.Framework.OptionsModel/ConfigurationBinder.cs
```diff
@@ -0,0 +1,253 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using Microsoft.Framework.ConfigurationModel;
+using Microsoft.Framework.Internal;
+
+namespace Microsoft.Framework.OptionsModel
+{
+    public class ConfigurationBinder
```
Eilon: This whole class pattern is a bit unusual... almost all the methods are static, except for the ctor and one method, and the only field is set once in the ctor and used once in that one method. There don't seem to be any cases where the instance `Bind` method is called more than once. The runtime doesn't seem to do it, the unit tests don't do it. Is that a scenario?

src/Microsoft.Framework.OptionsModel/ConfigurationBinder.cs
```diff
@@ -0,0 +1,253 @@
...
+                {
+                    if (typeInstance == null)
+                    {
+                        try
+                        {
+                            typeInstance = Activator.CreateInstance(type);
+                        }
+                        catch
+                        {
+                            // Cannot create the value so we cannot populate it
```
Eilon: A lot of empty or nearly-empty catch blocks here. Should this be logging debug info?

victorhurdugaci: There is already a bug on that (you logged it): https://github.com/aspnet/Options/issues/55

I didn't want to change the current functionality which ignores errors. I think we can address the empty catches as part of the other bug

HaoK: Might as well do it as part of a rewrite :). No sense waiting.

victorhurdugaci: On thing at a time :)

src/Microsoft.Framework.OptionsModel/ConfigurationBinder.cs
HaoK: Ok so then we need a better name. BindType doesn't really convey meaning. How is this different then the top level bind that takes an object?

src/Microsoft.Framework.OptionsModel/ConfigurationBinder.cs
Eilon: Oh yes https://github.com/aspnet/Options/issues/55 is a great bug :smile: assigned to you :smile: I'm ok doing in a separate PR.

src/Microsoft.Framework.OptionsModel/ConfigurationBinder.cs
HaoK: Yeah this is where I was going with my earlier feedback as well. I think this should all return to being a static class with all static methods. The class instance doesn't seem to be needed at all.

Eilon: I could see potentially being entirely non-static if it helps unit testing, but as it is right now it seems a bit contrived.

test/Microsoft.Framework.OptionsModel.Test/ConfigurationCollectionBindingTests.cs
victorhurdugaci: Added

src/Microsoft.Framework.OptionsModel/ConfigurationBinder.cs
victorhurdugaci: Changed to static

src/Microsoft.Framework.OptionsModel/ConfigurationBinder.cs
victorhurdugaci: Changed `BindObject` -> `BindObjectProperties`

src/Microsoft.Framework.OptionsModel/ConfigurationBinder.cs
```diff
@@ -0,0 +1,241 @@
...
+            propertyValue = BindType(property.PropertyType, propertyValue, configuration);
+
+            if (propertyValue != null && hasPublicSetter)
+            {
+                property.SetValue(propertyOwner, propertyValue);
+            }
+        }
+
+
+        private static object BindType(Type type, object typeInstance, IConfiguration configuration)
```
HaoK: Okay so this is really BindObject, and the old BindObject is really BindChildProperties then?

src/Microsoft.Framework.OptionsModel/ConfigurationBinder.cs
```diff
@@ -0,0 +1,241 @@
...
+        {
+            var configValue = configuration.Get(null);
```
HaoK: I've never seen this before, what does get passing in a null key to Configuration.Get do exactly?

victorhurdugaci: Gets the value of that configuration node. Currently, that's the only way to get the value

src/Microsoft.Framework.OptionsModel/ConfigurationBinder.cs
HaoK: Further more, should this just be the third public overload Bind? and you can probably switch the generic overload to use the new Bind

```
public static Bind<TModel>(IConfiguration) where TModel new() {
   return Bind(new Model(), typeof(TModel), config)
}

public static object Bind(object model, Type modelType, IConfiguration config)

HaoK: And do we actually need the typeless overload of Bind(object model, IConfiguration config)??  When do we not know the type?

src/Microsoft.Framework.OptionsModel/ConfigurationBinder.cs
HaoK: Ah ok, so that's how you extract the value from config leaf nodes...  That syntax just looks odd to me.  There's no other sugar like configuration.Value?

victorhurdugaci: Nope, that's the only way and for now it will stay like that - that was decided in the design meeting

src/Microsoft.Framework.OptionsModel/ConfigurationBinder.cs
victorhurdugaci: We don't support binding to unknown. It is like binding to `Dictionary<string, object>` which is already the same thing as querying the `Configuration` object.

src/Microsoft.Framework.OptionsModel/ConfigurationBinder.cs
HaoK: Oh well. Hopefully other people will complain too :)

victorhurdugaci: You should use options binding instead of configuration and not deal with that weirdness :)

HaoK: Well if we want to encourage that we could keep all of this internal and only expose it only via the configureoptions(IConfiguration) :)

vdangwal:
I am still not sure how to read an Array of objects in program.cs file using IConfigurationRoot or ConfigurationBuilder. Does anyone have example for the same?

PR #61, opened 2015-04-30
[Feedback] Use TypeDescriptor instead of Convert

suhasj:
@lodejard @divega @HaoK 

Porting this fix from Identity to use `TypeDescriptor.GetConverter` instead of `Convert` since `Convert` doesn't work for `Guid`. The if-def condition is added since the corresponding package doesn't support PCL. Some questions

- I was unable to get the if-def construct for PCL. I tried `NETPORTABLE`, `PCL` and `.NETPORTABLE,VERSION=V4.5,PROFILE=PROFILE7` with the last one not allowed. Can you point what is correct one ?
- The experience with this change is that, converting Guid is not possible for PCL projects. Do we add workaround for that ? In the discussion I had with @divega we had decided not to add this since PCL should eventually support Core CLR. Let me know if that has changed

### DIFF
--- a/src/Microsoft.Framework.OptionsModel/ConfigurationBinder.cs
+++ b/src/Microsoft.Framework.OptionsModel/ConfigurationBinder.cs
@@ -7,6 +7,7 @@ using System.Linq;
 using System.Reflection;
 using Microsoft.Framework.ConfigurationModel;
 using Microsoft.Framework.Internal;
+using System.ComponentModel;
 
 namespace Microsoft.Framework.OptionsModel
 {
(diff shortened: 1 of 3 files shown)


src/Microsoft.Framework.OptionsModel/ConfigurationBinder.cs
```diff
@@ -200,7 +201,12 @@ private static object CreateValueFromConfiguration(Type type, string value, ICon
                 }
                 else
                 {
+
+#if DNX451 || DNXCORE50 || NET45
+                    return TypeDescriptor.GetConverter(type).ConvertFromInvariantString(value);
+#else
                     return Convert.ChangeType(configuration.Get(null), type);
```
victorhurdugaci: If there are any cases in which we don't support the `TypeConverter`, as a workaround, we could special case those types. For example, for `Guid` we can call `Guid.Parse`

HaoK:
Code was moved to a different repo, closing

PR #62, opened 2015-04-30
Throw when configuration binding fails

victorhurdugaci:
Fixes: https://github.com/aspnet/Options/issues/55

I was inclined to add the configuration key in the `Data` property of the exception rather than the message. Let me know if you think that might be better.

Please review: @HaoK @Eilon


Eilon:
The Data property of exceptions is in fact a great place to store additional data that no one will ever see and will never get logged :smile: Just the exception message is fine. I'm borderline thinking it's not even necessary *there* either, but on the other hand I like it when errors give extra context as to where it's at when it all goes south.

src/Microsoft.Framework.OptionsModel/Resources.resx
```diff
@@ -0,0 +1,132 @@
...
+    <value>2.0</value>
+  </resheader>
+  <resheader name="reader">
+    <value>System.Resources.ResXResourceReader, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+  </resheader>
+  <resheader name="writer">
+    <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+  </resheader>
+  <data name="Error_CannotActivateAbstractOrInterface" xml:space="preserve">
+    <value>Cannot create instance of type '{0}' because it is either abstract or an interface (Configuration key: '{1}').</value>
```
Eilon: Maybe `Cannot blah blah blah blah. Configuration key: '{1}'` (one space after the first period, and no period at the very end)

In other words, make it be a "separate sentence", sort of like how argument exceptions do things.

src/Microsoft.Framework.OptionsModel/ConfigurationBinder.cs
```diff
@@ -26,18 +26,18 @@ public static void Bind(object model, IConfiguration configuration)
                 return;
             }
 
-            BindObjectProperties(model, configuration);
+            BindObjectProperties(model, configuration, null);
```
Eilon: Named param

Eilon:
Actually I take back much of what I said... I just took a closer look at the code and noticed we're now doing a lot of extra work just to have marginally better error messages. We're passing more parameters to every method call, as well as manufacturing new strings for every level of config. That's a lot of work for something that we don't even know is necessarily problematic or confusing.

I suggest simplifying this to avoid any extra work to calculate configuration keys. The rest of the errors are great, though.

victorhurdugaci:
I don't think the error message is just marginally better. It gives much more context.

Consider the two messages below:

```
Cannot convert `foo` to type `System.Int32`.

Cannot convert `foo` to type `System.Int32`. Configuration key: `Locations:Address:0:ZipCode`.
```

The first message tells you that something is wrong but it doesn't tell you where the error is. It is like getting an `unexpected token {` error from a json parser without telling you where in the file it is not expected.

Also, `foo` can appear multiple times in the configuration. Which `foo` is the broken one?

Eilon:
I'm just not a fan of doing a fair bit of extra work for something that's only occasionally useful. If the invalid value is displayed it's usually trivial to find it in the config file.

As far as being useful, I think it'd be far more useful to show which file / source the error is from, but our config system doesn't support that type of introspection at all.

If this information was all readily available and we were just showing it, I think that'd be fine. But we're doing extra computations, allocations, etc. "just in case" there's an error.

HaoK:
Yeah I agree with Eilon. When I first saw this change my immediate reaction was eew gross this doesn't seem worth it. So if this isn't super useful it should go. Passing the config key everywhere seems too much

Eilon:
@HaoK BTW it's not just *passing* it, it's *creating* it (manufacturing new strings, etc.).

HaoK:
Maybe if this is a common scenario what's missing is a config API like config.GetAbsoluteNodePath which would know how to do the traversal and print out "root:key1:key2".  But yeah that seems a bit much to build that manually for exception messages

victorhurdugaci:
Updated - removed the context

src/Microsoft.Framework.OptionsModel/Resources.resx
```diff
@@ -0,0 +1,132 @@
...
+    <value>System.Resources.ResXResourceReader, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+  </resheader>
+  <resheader name="writer">
+    <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+  </resheader>
+  <data name="Error_CannotActivateAbstractOrInterface" xml:space="preserve">
+    <value>Cannot create instance of type '{0}' because it is either abstract or an interface.</value>
+  </data>
+  <data name="Error_FailedBinding" xml:space="preserve">
+    <value>Failed to convert '{0}' to type '{1}'. See InnerException for details.</value>
```
Eilon: I think that `See InnerException for details.` is unnecessary - the call stack should will show all the chained exceptions anyway.

src/Microsoft.Framework.OptionsModel/ConfigurationBinder.cs
```diff
@@ -141,16 +154,13 @@ private static void BindDictionary(object dictionary, Type iDictionaryType, ICon
...
-                    var item = BindType(valueType, null, keyConfiguration);
-                    if (item != null)
-                    {
-                        addMethod.Invoke(dictionary, new[] { keyProperty.Key, item });
-                    }
-                }
-                catch
+                var item = BindType(
+                    valueType,
+                    null,
```
Eilon: Named param for this one?

src/Microsoft.Framework.OptionsModel/ConfigurationBinder.cs
```diff
@@ -172,7 +182,10 @@ private static void BindCollection(object collection, Type iCollectionType, ICon
 
                 try
                 {
-                    var item = BindType(itemType, null, keyConfiguration);
+                    var item = BindType(
+                        itemType,
+                        null,
```
Eilon: Named param

test/Microsoft.Framework.OptionsModel.Test/ConfigurationBinderExceptionTests.cs
```diff
@@ -0,0 +1,144 @@
...
+namespace Microsoft.Framework.OptionsModel.Test
+{
+    public class ConfigurationBinderExceptionTests
+    {
+        [Fact]
+        public void ExceptionWhenTryingToBindToInterface()
+        {
+            var input = new Dictionary<string, string>
+            {
+                {"ISomeInterfaceProperty:Subkey", "x"}
```
Eilon: I think the default VS formatting puts a space after the `{` and before the `}` no? If so, please format this so that the next person won't have a bunch of space-only changes.

Eilon: In other words, <kbd>Ctrl</kbd>+<kbd>K</kbd> <kbd>Ctrl</kbd>+<kbd>D</kbd>.

victorhurdugaci: That key combination is only for the C# dev environment ;)

victorhurdugaci: And, no, VS is not adding spaces. Actually, it doesn't format that at all. If I add spaces it leaves them there, if there are no spaces it is not adding them

Eilon: Hmm ok!

victorhurdugaci:
Updated with named params

Eilon:
:shipit: after fixing the xproj.

Issue #60, opened 2015-04-30
Add support for binding property to IConfiguration


davidfowl:
In omnisharp I want to have a configuration like this:

```JSON
{
   "msbuild": {
      "toolsVersion": "14.0",
       "properties": {
           "RandomProperty": "1",
           "VisualStudioVersion": "14.0"
        }
   }
}
```

I would like to express my options like so:

```C#
public class MSBuildOptions
{
     public string ToolsVersion { get; set; }
     public IConfiguration Properties { get; set; }
}
```

That way I would be able to express both the latebound and strongly typed properties on the same model object.

/cc @glennc @muratg @Eilon @lodejard

Eilon:
This is like a combination of model binding and tag helpers :smile: 

We do support binding to dictionaries now with @victorhurdugaci 's change, no? Would that work here?

davidfowl:
If I have a complex structure this gives me more flexibility

Eilon:
True, but in the specific case you gave I think it's supported quite well. So, backlog?

davidfowl:
Sure

davidfowl:
Is it string,string?

victorhurdugaci:
If you have only one level of nesting like in your example, `Dictionary<string, string>` will work. If you have multiple levels (e.g. `VisualStudioVersion` being a complex object), we don't support it that today and you have to use `IConfiguration`.

Eilon:
I think it does all kinds of conversions for you... https://github.com/aspnet/Options/blob/dev/test/Microsoft.Framework.OptionsModel.Test/ConfigurationCollectionBindingTests.cs

It definitely seems to support `Dictionary<string, string>` though I'm not sure if `TValue` can be a simple type such as `int`, though I do see it can also be a complex type.

victorhurdugaci:
`Dictionary<string, T>` is supported as long as all Ts are the same type. T can be int.

Eilon:
Ah don't think I saw a test for that... did I miss it? If it's not there, can we quickly add one?

victorhurdugaci:
Sure, I can add one tomorrow

victorhurdugaci:
So, is this backlog or not?

Eilon:
@davidfowl how passionate are you about this? For me this seems like super backlog. We can bring this back at any time.

HaoK:
Config binder has moved to config and there's a similar bug here: https://github.com/aspnet/Configuration/issues/274

PR #63, opened 2015-05-01
Tests for config collection binding with non string values

victorhurdugaci:
@Eilon these are tests that you asked to add [here](https://github.com/aspnet/Options/issues/60#issuecomment-98031781)


Eilon:
:shipit: !

@davidfowl FYI.

PR #65, opened 2015-05-18
Move ConfigurationBinder to Configuration

HaoK:
Options side of https://github.com/aspnet/Configuration/pull/205

Note: we can't depend on Configuration Abstractions unless ConfigurationBinder lives there

cc @divega


brockallen:
Did this actually happen? I'm in RC1 and I still see these extension methods in the Options NuGet, not the core Configuration NuGet.

HaoK:
Yeah early on a08ccc77e6a090a357a144167b30688f75626d93

brockallen:
Hmm, ok... In RC1 I had to include the Options NuGet to get the APIs/extension methods. I'll go back and double check...

brockallen:
Ok, I see what went wrong in my sample. I thought the extension was coming from Options, but when I pulled in Options it also pulled in the Config.Binder assembly. So yes, the extension method was really in the Binder helper (and not options). Thx for being my cardboard cutout.

PR #68, opened 2015-06-16
using statement duplicated

ryan1234:
Trying to debug some code, looking for inspiration and noticed a duplicated using.


Eilon:
Nice :smile:

PR #71, opened 2015-07-01
Add repository information to packages

victorhurdugaci:
Adding repository information to the packages that are published.

@davidfowl @troydai

### DIFF
--- a/src/Microsoft.Framework.OptionsModel/project.json
+++ b/src/Microsoft.Framework.OptionsModel/project.json
@@ -6,6 +6,10 @@
     "Microsoft.Framework.DependencyInjection.Abstractions": "1.0.0-*",
     "Microsoft.Framework.NotNullAttribute.Sources": { "type": "build", "version": "1.0.0-*" }
   },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/aspnet/options"
+  },
   "frameworks": {
     "net45": { },
     "dnx451": { },


troydai:
:shipit:

src/Microsoft.Framework.OptionsModel/project.json
```diff
@@ -6,6 +6,10 @@
     "Microsoft.Framework.DependencyInjection.Abstractions": "1.0.0-*",
     "Microsoft.Framework.NotNullAttribute.Sources": { "type": "build", "version": "1.0.0-*" }
   },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/aspnet/options"
```
Eilon: So how does this really work? It doesn't need a branch or tag or something?

victorhurdugaci: Nope. It picks everything else when it builds. But you must build from a folder that's a git repo

Sent from my phone
________________________________

Eilon: So does this work out of the box? I read https://github.com/aspnet/dnx/issues/1901 but I don't see how this info here is enough for it to completely work.

victorhurdugaci: If you add the repository information to you project.json it works out of the box. All it needs in addition to the repo type and the url is the commit ID and the folder, relative to the repo root, where the project it.

Eilon: Right, so how does the commit ID make it in? Does our build do that?

victorhurdugaci: `dnu pack` invokes `git` to get the commit id

Eilon: Ah cool, I see that in the `repo.json` in the generated NuGet package... cool!

PR #77, opened 2015-07-28
IOptions.Options => Value, RIP IOptions.Order

HaoK:
So many less overloads...yay

cc @divega


HaoK:
Note: I still need to fix MVC/Security before this will be merged

src/Microsoft.Framework.OptionsModel/OptionsManager.cs
```diff
@@ -1,17 +1,14 @@
...
-using Microsoft.Framework.Internal;
 
 namespace Microsoft.Framework.OptionsModel
 {
     public class OptionsManager<TOptions> : IOptions<TOptions> where TOptions : class,new()
     {
-        private object _mapLock = new object();
-        private Dictionary<string, TOptions> _namedOptions = new Dictionary<string, TOptions>(StringComparer.OrdinalIgnoreCase);
+        private TOptions _options;
         private IEnumerable<IConfigureOptions<TOptions>> _setups;
```
divega: readonly?

src/Microsoft.Framework.OptionsModel/OptionsManager.cs
```diff
@@ -19,39 +16,22 @@ public OptionsManager(IEnumerable<IConfigureOptions<TOptions>> setups)
...
+            get
             {
-                lock (_mapLock)
+                if (_options == null)
                 {
-                    if (!_namedOptions.ContainsKey(name))
-                    {
-                        _namedOptions[name] = Configure(name);
-                    }
+                    _options = _setups == null
```
divega: Is _setups ever null? Sorry, can't remember DI behaviors.

divega:
:shipit: with few comments.

HaoK:
@divega What do you think about a new abstract base class

```
public abstract class Options<TOptions> : IOptions<TOptions> {
  IOptions.Value { get { return this; } }
}
```

HaoK:
Will need to do this a fair bit, might as well save the mechanical nature and provide a base class to save a few lines of code for everyone.

PinpointTownes:
Why not a new `Options<T>` class acting like a wrapper around a singleton instance?

https://github.com/PinpointTownes/Security/commit/26b511f3d294f4b9545e1fbc57e4ef5226672118#diff-e1b2bedb31c9059949ac46787b48b1b6R10

HaoK:
c623fbc3c6a8b037ed191da950587d6fa884bcac

PinpointTownes:
Noooooooooooo, I'll miss you, named options :cry:

PR #79, opened 2015-08-06
Microsoft.Framework.OptionsModel.Test -> Microsoft.Framework.OptionsModel.Tests

hishamco:
According to the guidelines https://github.com/aspnet/Home/wiki/Engineering-guidelines#solution-and-project-folder-structure-and-naming, we should rename the test project to ```Microsoft.Framework.OptionsModel.Tests``` instead of ```Microsoft.Framework.OptionsModel.Test```


Eilon:
See https://github.com/aspnet/EventNotification/pull/12#issuecomment-128427434.

PR #99, opened 2015-11-11
Add support for reloadable options

HaoK:
First iteration of reloadable options aka IOptionsWatcher see: 
https://github.com/aspnet/specs/blob/master/notes/reloadable_options.md

- Split Configuration specific functionality into new assembly: Options.ConfigurationExtensions
- IOptionsWatcher is the service that lets you `Watch(Action<TOptions> callback)`
- IOptionsChangeTracker is the service fires the watcher (i.e. the ConfigChangeTracker): 

Usage example:
```
        public class Controller : IDisposable
        {
            IDisposable _watcher;
            FakeOptions _options;

            public Controller(IOptionsWatcher<FakeOptions> watcher)
            {
                _watcher = watcher.Watch(o => _options = o);
            }

            public void Dispose()
            {
                _watcher?.Dispose();
            }

            public string Message => _options?.Message;
        }

            // Startup glue to listen to configuration changes to trigger reload
            services.AddOptions().TrackChanges<FakeOptions>(config);

```

cc @divega @rynowak @lodejard


src/Microsoft.Extensions.Options.ConfigurationExtensions/ConfigurationChangeTracker.cs
```diff
@@ -0,0 +1,33 @@
...
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.Primitives;
+
+namespace Microsoft.Extensions.OptionsModel
+{
+    public class ConfigurationChangeTracker<TOptions> : IOptionsChangeTracker<TOptions>
+    {
+        private IConfiguration _config;
+
+        public ConfigurationChangeTracker(IConfiguration config)
```
rynowak: Should this be `IConfigurationRoot`?

src/Microsoft.Extensions.Options.ConfigurationExtensions/project.json
```diff
@@ -0,0 +1,46 @@
+{
+  "version": "1.0.0-*",
+  "compilationOptions": {
+    "warningsAsErrors": true,
+    "keyFile": "../../tools/Key.snk"
+  },
+  "description": "Provides a strongly typed way of specifying and accessing settings using dependency injection.",
```
rynowak: Description needs some love

src/Microsoft.Extensions.OptionsModel/IOptionsWatcher.cs
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Extensions.OptionsModel
+{
+    public interface IOptionsWatcher<out TOptions>
```
rynowak: Why is this `IOptionsWatcher` instead of `IReloadableOptions`?

src/Microsoft.Extensions.OptionsModel/OptionsServiceCollectionExtensions.cs
```diff
@@ -21,6 +20,8 @@ public static IServiceCollection AddOptions(this IServiceCollection services)
             }
 
             services.TryAdd(ServiceDescriptor.Singleton(typeof(IOptions<>), typeof(OptionsManager<>)));
+            // REVIEW: what do we want the lifetime to be?  TODO: make this scoped and try to automatically do IDisposable tracking
```
rynowak: That's an interesting idea. I don't think you could completely solve leaks by making it scoped - because components might request it and each have different lifetimes.

src/Microsoft.Extensions.OptionsModel/OptionsWatcher.cs
```diff
@@ -0,0 +1,60 @@
...
+using System.Linq;
+using Microsoft.Extensions.Primitives;
+
+namespace Microsoft.Extensions.OptionsModel
+{
+    public class OptionsWatcher<TOptions> : IOptionsWatcher<TOptions> where TOptions : class, new()
+    {
+        private TOptions _options;
+        private IEnumerable<IConfigureOptions<TOptions>> _setups;
+        private IOptionsChangeTracker<TOptions> _tracker;
```
rynowak: these can be made readonly

src/Microsoft.Extensions.OptionsModel/OptionsWatcher.cs
```diff
@@ -0,0 +1,60 @@
...
+        {
+            return _setups == null
+                ? new TOptions()
+                : _setups.Aggregate(new TOptions(),
+                                    (options, setup) =>
+                                    {
+                                        setup.Configure(options);
+                                        return options;
+                                    });
+
```
rynowak: This is a mess to read

src/Microsoft.Extensions.OptionsModel/project.json
```diff
@@ -6,9 +6,8 @@
   },
   "description": "Provides a strongly typed way of specifying and accessing settings using dependency injection.",
   "dependencies": {
-    "Microsoft.Extensions.Configuration.Abstractions": "1.0.0-*",
-    "Microsoft.Extensions.Configuration.Binder": "1.0.0-*",
-    "Microsoft.Extensions.DependencyInjection.Abstractions": "1.0.0-*"
+    "Microsoft.Extensions.DependencyInjection.Abstractions": "1.0.0-*",
+    "Microsoft.Extensions.Primitives": "1.0.0-*"
```
rynowak: It's really nice to drop the configuration dependency here

src/Microsoft.Extensions.OptionsModel/IOptionsChangeTracker.cs
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.Extensions.Primitives;
+
+namespace Microsoft.Extensions.OptionsModel
+{
+    public interface IOptionsChangeTracker<out TOptions>
```
rynowak: https://github.com/aspnet/specs/blob/master/notes/reloadable_options.md#proposed-api-changes

This also deviates from what we had spec'ed out. Why this approach over the other? I think it's probably simpler this way but curious.

src/Microsoft.Extensions.OptionsModel/OptionsWatcher.cs
```diff
@@ -0,0 +1,60 @@
...
+        }
+
+        public IDisposable Watch(Action<TOptions> watcher)
+        {
+            return ChangeTokenHelper.OnChange(_tracker.GetChangeToken, () =>
+            {
+                // Recompute the options: REVIEW is there thread safety issues here?
+                _options = SetupValue();
+                watcher(_options);
+            });
```
rynowak: This isn't right because `CurrentValue` will never change.

Take a look at the sequence diagram https://github.com/aspnet/specs/blob/master/notes/reloadable_options.md#the-surrounding-object-needs-to-be-prepared-to-deal-with-that-case

src/Microsoft.Extensions.OptionsModel/OptionsServiceCollectionExtensions.cs
divega: Fair point. But it is a shame because it would make the most common case much easier to deal with (i.e. no need to implement IDisposable in your controllers).

src/Microsoft.Extensions.OptionsModel/IOptionsChangeTracker.cs
divega: Because it was just a spec? :smile:

src/Microsoft.Extensions.OptionsModel/OptionsServiceCollectionExtensions.cs
lodejard: If you make it scoped wouldn't it re-initialize the TOptions from scratch per-request?

src/Microsoft.Extensions.Options.ConfigurationExtensions/ConfigurationChangeTracker.cs
HaoK: Diego suggested we add Root to the IConfiguration interface so we don't have to pass in the root here.

src/Microsoft.Extensions.OptionsModel/IOptionsWatcher.cs
HaoK: Are you in love with that name?  Not a huge fan so I was trying some others on for size... IOptionsMonitor also has the same semantic meaning as Watcher.  I view this class as the similar to FileMonitor/Watcher, as you get notified when your options change

rynowak: Personally, I don't really mind the name change way. If everyone's happy I'm happy.

src/Microsoft.Extensions.OptionsModel/OptionsServiceCollectionExtensions.cs
HaoK: Yes it would reinitialize the options every request, as transient, its worse than that as you get a new TOptions every time.  So I guess that answers the lifetime question, this should be singleton just like the other IOptionsManager then.

src/Microsoft.Extensions.OptionsModel/OptionsWatcher.cs
HaoK: yeah its ugly, linq hell. Let me find some lipstick...

src/Microsoft.Extensions.OptionsModel/IOptionsChangeTracker.cs
HaoK: Yeah after banging on it for a bit, this felt nicer to me too

src/Microsoft.Extensions.OptionsModel/OptionsWatcher.cs
HaoK: So it was unclear to me what the behavior of CurrentValue was suppose to be.  Diego thought it should cache the value, and only be updated whenever things change.  My initial iteration I had it basically run the Setups every time its called to truly compute the 'CurrentValue'.  As things are currently implemented, the consumers are passed the new value directly in their `Action<TOptions>`.  So today, the question is what should CurrentValue return when they hit it from IOptionsWatcher directly.  The last value returned by a change, or the initial value when the object was constructed?  Or is this all way off base :)

rynowak: I think my comment was just wrong. Let me take another look at this.

src/Microsoft.Extensions.OptionsModel/IOptionsWatcher.cs
HaoK: 3 name votes left, Ryan abstains :)

src/Microsoft.Extensions.Options.ConfigurationExtensions/ConfigurationChangeTracker.cs
rynowak: Is there a scenario I'm missing out on here? Why wouldn't you pass the root?

src/Microsoft.Extensions.OptionsModel/OptionsWatcher.cs
HaoK: Basically we would be rerunning the DI configured options so we'd pick up potentially any changes from Config bindings, or any other setups that potentially change behavior from the current services in DI.

src/Microsoft.Extensions.OptionsModel/OptionsWatcher.cs
```diff
@@ -0,0 +1,60 @@
...
+namespace Microsoft.Extensions.OptionsModel
+{
+    public class OptionsWatcher<TOptions> : IOptionsWatcher<TOptions> where TOptions : class, new()
+    {
+        private TOptions _options;
+        private IEnumerable<IConfigureOptions<TOptions>> _setups;
+        private IOptionsChangeTracker<TOptions> _tracker;
+
+        // REVIEW: IEnumerable monitors?
+        public OptionsWatcher(IEnumerable<IConfigureOptions<TOptions>> setups, IOptionsChangeTracker<TOptions> tracker)
```
rynowak: Yeah, it sounds like you'd need to support multiple change trackers here. Nothing in the API stops you from wiring it up that way.

src/Microsoft.Extensions.Options.ConfigurationExtensions/ConfigurationChangeTracker.cs
HaoK: Well we bind against IConfiguraiton usually for options, felt more convenient to be able to take any part of config (since internally they know how to get to the root anyways).  This way you can be passed the relevant section (if we wanted to have an overload for Bind + TrackChanges in one call, i.e. `Configure<MyOptions>(config.Get("My", trackChanges: true)`

src/Microsoft.Extensions.OptionsModel/OptionsWatcher.cs
rynowak: I think `CurrentValue` is coded correctly right now. Also there's no threading issue, you don't register for notifications again until after setting `_options`.

> As things are currently implemented, the consumers are passed the new value directly in their `Action<TOptions>`

This matches my expectations

rynowak: FYI if we add multiple change trackers then this will have to have some kind of mutex.

src/Microsoft.Extensions.Options.ConfigurationExtensions/ConfigurationChangeTracker.cs
rynowak: Ok, that makes much more sense why this would be useful.

src/Microsoft.Extensions.OptionsModel/OptionsWatcher.cs
HaoK: So we'd basically wire up an OnChange for each tracker that fires the Action, and keep a list of the IDisposables and return a new one that wraps disposing all of the ones we created?

HaoK:
Updated and added an OptionsCache helper class that uses LazyInitializer to guard against concurrency and updated both OptionsWatcher/Manager to share this.  Will add support for multiple Trackers next...

HaoK:
Updated added support for multiple trackers.  Question, do we need some way to Unwatch/unregister for notifications?  Dispose doesn't appear to stop notifications from config, I was able to verify dispose does call through in a unit test, but at least for the ConfigurationReloadToken disposing the watch does not prevent the event from firing later.

divega:
>  do we need some way to Unwatch/unregister for notifications

Flowing `Dispose()` all the way to the `CancellationTokenRegistration` should be the way to do this. Are you sure that we are not missing anything? Maybe there is a bug in `CancellationToken`?

HaoK:
Okay looks like there's a subtle bug specifically with the config.Reload apis:

This works, where if you dispose before the first reload, the callback is never fired.
```
            var cleanup = watcher.Watch(o => updatedMessage = o.Message);

            cleanup.Dispose();
            config.Reload();
            // Correctly does not fire...
```

But firing the callback once, makes the first disposable not be live anymore, so there's no way to 'dispose the new change token' as we can't get a handle to that with the current API

```
            var cleanup = watcher.Watch(o => updatedMessage = o.Message);
            config.Reload();
            // message is now updated
            cleanup.Dispose();
            config.Reload();
            // I would expect watch not to get fired...  It fires... (because cleanup was already used)
```

src/Microsoft.Extensions.OptionsModel/OptionsCache.cs
```diff
@@ -0,0 +1,47 @@
...
+        }
+
+        private TOptions CreateOptions()
+        {
+            var result = new TOptions();
+            return _setups?.Aggregate(result, (options, setup) =>
+            {
+                setup.Configure(options);
+                return options;
+            }) ?? result;
```
rynowak: Is this really better than a for loop and an if?

rynowak:
> So you expect Watch to immediately fire when called? And also on future changes?

Yes.

HaoK:
The bug is most likely in the ghetto IDisposable wrapper I added to handle multiple trackers.

```
            var disposable = new ChangeTrackerDisposable();
            foreach (var tracker in _trackers)
            {
                disposable.Disposables.Add(ChangeToken.OnChange(tracker.GetChangeToken, () =>
                {
                    // Recompute the options before calling the watchers
                    _optionsCache = new OptionsCache<TOptions>(_setups);
                    watcher(_optionsCache.Value);
                }));
            }
            return disposable;
```

HaoK:
Just curious why do we want Watch to immediately call the action?  Watch isn't the best name if we want that behavior.  They could replicate that behavior by calling their method themselves first too right?

```
   myAction(watch.CurrentValue)
   watcher.Watch(myActiion)
```

Is there a reason to bake that into the method?

HaoK:
Hrm so the leak appears to be in the ChangeToken helper:

```
            Action<object> callback = null;
            callback = (s) =>
            {
                var t = changeTokenProducer();
                changeTokenConsumer((TState)s);
                t.RegisterChangeCallback(callback, s);
            };

            return changeTokenProducer().RegisterChangeCallback(callback, state);
```

We don't return the t.RegisterChangeCallback that's inside of the callback, I fixed the issue by not using the extension method for the Options watcher for now...

HaoK:
Updated with latest naming suggestions from @divega 
IOptionsMonitor (aka IReloadableOptions/IOptionsWatcher)
IOptionsChangeTokenSource (aka IOptionsChangeTracker)

Stopped using ChangeToken.OnChange for now, since we need to remember each call to RegisterChangeCallback for cleanup.

There are no longer any issues I'm aware of, so I think this is ready for final review

src/Microsoft.Extensions.OptionsModel/OptionsMonitor.cs
```diff
@@ -0,0 +1,72 @@
...
+                    // registering. This prevents us from possible having two change updates to process concurrently.
+                    //
+                    // If the token changes after we take the token, then we'll process the update immediately upon
+                    // registering the callback.
+                    var token = source.GetChangeToken();
+
+                    // Recompute the options before calling the watchers
+                    _optionsCache = new OptionsCache<TOptions>(_setups);
+                    listener(_optionsCache.Value);
+                    disposable.Disposables.Add(token.RegisterChangeCallback(callback, s));
```
rynowak: Will this list ever get cleared or is this going to grow unbounded?

rynowak: We might need to close over another variable here holding the 'previous' disposable.

HaoK: Yeah I noticed that too, we never prune from the list, so it will just grow forever in a singleton no less :)

HaoK: Actually i guess its not a singleton, but each list will continue to grow as long as it keeps firing.

HaoK: Updated, to track the previousSubscription and prune the list, I made the disposable internal and updated the test verify that the count doesn't keep growing now

test/Microsoft.Extensions.OptionsModel.Test/OptionsMonitorTest.cs
```diff
@@ -0,0 +1,220 @@
...
+
+            var cleanup = monitor.OnChange(o => updatedMessage = o.Message) as ChangeTrackerDisposable;
+
+            config.Reload();
+            Assert.Equal("2", updatedMessage);
+
+            // Verify old watch is changed too
+            Assert.Equal("2", monitor.CurrentValue.Message);
+
+            Assert.Equal(1, cleanup.Disposables.Count);
```
HaoK: This was previously growing each time OnChange was fired

rynowak:
:shipit: from me

HaoK:
Ok now time to make Diego/Lou happy with the config PR that this depends on :)

src/Microsoft.Extensions.Options.ConfigurationExtensions/OptionsServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,50 @@
...
+            if (config == null)
+            {
+                throw new ArgumentNullException(nameof(config));
+            }
+
+            services.ConfigureOptions(new ConfigureFromConfigurationOptions<TOptions>(config));
+            return services;
+        }
+
+        public static IServiceCollection Configure<TOptions>(this IServiceCollection services, IConfiguration config, bool trackConfigChanges)
```
HaoK: trackChanges ?

HaoK:
ec13249a35bbf1d4a8a381a9697965cc1208fe82

Issue #98, opened 2015-11-06
DateTime ConvertFromInvariantString test fail ( as expected )


devel0:
Running unit tests I got fail at  `Configure_GetsNullableOptionsFromConfiguration`  test for given test datetime, this is expected cause the test try to ConvertFromInvariantString at ConfigurationBinder.cs (ReadValue) where passed input string is generated by the OptionsTest.cs(Configure_GetsNullableOptionsFromConfiguration_Data) where a datetime string is generated using current culture instead of the invariant one.
```{ nameof(NullableOptions.MyNullableDateTime), new DateTime(1995, 12, 31).ToString(CultureInfo.CurrentCulture.DateTimeFormat.ShortDatePattern) }```

HaoK:
@divega what behavior do we want here?  Indeed we are always converting from Invariant in the binder, should we try both with Invariant first and then a normal ConvertFromString if the first fails?

MichaCo:
I'd say it is not a good idea to parse with the current culture. Instead, this test should be fixed with setting the culture correctly.

What would happen if you have a date string formatted with the `de-DE` culture in your configuration and deploy that to production where the culture is `en-US`? Yup it would fail...
*runs on my machine* issue :crying_cat_face:

HaoK:
Yeah I guess what we have is fine, its just a test issue, and invariant string format is the only supported one

HaoK:
@divega is it expected that we need to support running our tests outside of en-US?  Either way the priority of this should be low as its not causing any issues unless you run tests with a different culture... Moving to backlog

HaoK:
758e11ecee0c30e12d3498182a799148d7d45a8c

divega:
Also, I have filed https://github.com/aspnet/Configuration/issues/348 to follow up on what behavior we want on the product re culture.

PR #101, opened 2015-12-08
Fix test to always use invariant culture

HaoK:
cc @divega

### DIFF
--- a/test/Microsoft.Extensions.OptionsModel.Test/OptionsTest.cs
+++ b/test/Microsoft.Extensions.OptionsModel.Test/OptionsTest.cs
@@ -5,7 +5,6 @@ using System;
 using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
-using System.Reflection;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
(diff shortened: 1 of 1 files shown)


divega:
@HaoK could you briefly explain how this is going to help? I get that we are now producing theory data using invariant culture, I just want to understand how we are consuming this in the tests and how the culture has an impact.

HaoK:
There was a test bug where this test fails when running with current culture is not en-us. This fixes that as the underlying code in config binder only converts from invariant culture anyways.

divega:
:shipit:

> This fixes that as the underlying code in config binder only converts from invariant culture anyways.

I will file a bug on that :smile:

MichaCo:
There is also the Microsoft.AspNet.Testing.ReplaceCultureAttribute which temporarily sets the current thread culture.

Regarding the binder using invariant only; we discussed that in #98. 
I think it is dangerous to assume that the development and maybe production/other environments run with the same system culture. 
And in those cases with current culture you might get different results reading the same configuration 

But extending the binder to take the culture optionally, sounds like a good idea!

HaoK:
758e11ecee0c30e12d3498182a799148d7d45a8c

PR #103, opened 2015-12-15
Rename OptionsModel => Options

HaoK:
cc @divega


test/Microsoft.Extensions.Options.Test/OptionsTest.cs
```diff
@@ -0,0 +1,306 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
```
divega: Curious, is this file new or a rename/move? It seems that the diff is missing something.

divega:
This seems to be a good change to me and I think we should take it as long as stakeholders are ok with the impact of the breaking change (I personally think it is ok). Will it require changes to templates or other places? We should do an announcements post.

cc @Eilon @DamianEdwards

test/Microsoft.Extensions.Options.Test/OptionsTest.cs
HaoK: Its a rename/move, I stash/unstashed since I made the initial changes in the wrong branch which might have confused git

HaoK:
Yeah this one will require changes in a ton of places, but holidays are the best time to push changes like these threw if we decide to do this :)

Eilon:
Aside from being a hassle to everyone, is there any reason *not* to take it? I certainly prefer the new name.

HaoK:
I'll roll out this change next week then when hopefully things are quiet

HaoK:
60130801e772a6cc811c182c6d260c44e4559d3b

PR #104, opened 2015-12-18
Add TryConfigure method

henkmollema:
I'd like to be able to only configure options when they have not been configured already. E.g. when a framework sets default options but they can be overridden by the end-user.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,4 @@ nuget.exe
 *.ipch
 *.sln.ide
 project.lock.json
+.vs/
--- a/Options.sln
+++ b/Options.sln
@@ -1,11 +1,22 @@
-
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 14
-VisualStudioVersion = 14.0.21722.1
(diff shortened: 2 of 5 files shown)


divega:
@HaoK I think we talked about taking this but that it needed to be re-implemented based on `TryAdd()` and obviously rebased with all the renaming churn :sweat:. Can you take a look?

henkmollema:
@divega do you mean something like this?

```csharp
services.TryAddSingleton<IConfigureOptions<TOptions>>(new ConfigureOptions<TOptions>(setup));
// or
services.TryAdd(ServiceDescriptor.Singleton<IConfigureOptions<TOptions>>(new ConfigureOptions<TOptions>(setup)));
```

divega:
Whatever works. If the code reads similar to the non-try version on https://github.com/aspnet/Options/blob/dev/src/Microsoft.Extensions.Options.ConfigurationExtensions/OptionsServiceCollectionExtensions.cs the better. The main issue is that the method that was originally in the PR doesn't exist anymore.

Issue #106, opened 2016-01-05
Add Options<T>


davidfowl:
Would act as an `IOptions<T>` wrapper for some T. Would make a convenient way to create an `IOptions<T>` from any T. We could also have a factory method (/cc @divega) that makes one of these from a T i.e. `IOptions<T> Options.Create<T>(T options)`.

HaoK:
Would this would be instead of the old pattern of MyOptions implementing `IOptions<MyOptions>`?

davidfowl:
Yep! It's something @PinpointTownes suggested https://github.com/aspnet/Options/issues/105#issuecomment-168866368.

davidfowl:
I updated the sample in B to show the usage.

HaoK:
Ok yeah that looks much better, it was pretty gross making everything implement IOptions before...  Maybe we should name the generic Options class OptionsWrapper so people don't think its a base class...which is what the name implies...

davidfowl:
PR out for review.

HaoK:
Do we want to consider an overload of `Options.Create<T>(Action<T> configure)` just as pure sugar?

davidfowl:
Hmmmmmmmm

davidfowl:
```C#
app.UseFoo(Options.Create(options =>
{
    options.Value = 10;
});
```

```C#
var options = Options.Create(opt =>
{
  opt.Value = 10;
});

app.UseFoo(options);
```

HaoK:
Yeah well this way you give people the option to do either one, so everyone is happy

Issue #100, opened 2015-12-07
Consider merging IOptionsWatcher + IOptions


HaoK:

shirhatti:
`Options` is going away

PR #107, opened 2016-01-05
Added a convinence wrapper for making IOptions<T> instances

davidfowl:
- Useful for passing poco `TOptions` to APIs that take `IOptions<TOptions>`

#106

/cc @HaoK @divega @DamianEdwards

### DIFF
--- a/src/Microsoft.Extensions.Options/Options.cs
+++ b/src/Microsoft.Extensions.Options/Options.cs
@@ -1,0 +1,10 @@
+﻿namespace Microsoft.Extensions.Options
+{
+    public static class Options
+    {
+        public static IOptions<TOptions> Create<TOptions>(TOptions options) where TOptions : class, new()
+        {
+            return new OptionsWrapper<TOptions>(options);
+        }
+    }
+}
--- a/test/Microsoft.Extensions.Options.Test/OptionsTest.cs
+++ b/test/Microsoft.Extensions.Options.Test/OptionsTest.cs
@@ -302,5 +302,45 @@ namespace Microsoft.Extensions.Options.Tests
             Assert.Collection(expectedValues, assertions.ToArray());
         }
 
+        [Fact]
+        public void Options_StaticCreateCreateMakesOptions()
+        {
+            var options = Options.Create(new FakeOptions
+            {
+                Message = "This is a message"
+            });
+
+            Assert.Equal("This is a message", options.Value.Message);
+        }
+
+        [Fact]
+        public void OptionsWrapper_MakesOptions()
+        {
+            var options = new OptionsWrapper<FakeOptions>(new FakeOptions
+            {
+                Message = "This is a message"
+            });
+
+            Assert.Equal("This is a message", options.Value.Message);
+        }
+
+        [Fact]
+        public void Options_CanOverrideForSpecificTOptions()
+        {
+            var services = new ServiceCollection().AddOptions();
+
+            services.Configure<FakeOptions>(options =>
+            {
+                options.Message = "Initial value";
+            });
+
+            services.AddSingleton(Options.Create(new FakeOptions
+            {
+                Message = "Override"
+            }));
+
+            var sp = services.BuildServiceProvider();
+            Assert.Equal("Override", sp.GetRequiredService<IOptions<FakeOptions>>().Value.Message);
+        }
     }
 }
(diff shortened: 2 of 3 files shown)


davidfowl:
Issues may arise if users register `IOptions<T>` in the container. It'll overwrite the `OptionsManager<T>` implementation.

HaoK:
I think it might be okay as they would likely register `IOptions<MyOptions>` not `IOptions<>` which would only override a specific option type perhaps?

davidfowl:
@HaoK True.

src/Microsoft.Extensions.Options/Options.cs
```diff
@@ -0,0 +1,20 @@
...
+{
+    public static class Options
+    {
+        public static IOptions<TOptions> Create<TOptions>(TOptions options) where TOptions : class, new()
+        {
+            return new Options<TOptions>(options);
+        }
+    }
+
+    public class Options<TOptions> : IOptions<TOptions> where TOptions : class, new()
```
HaoK: I think we should name this something less generic, as it currently stands `Options<TOptions>` might lead people to think this is the base class for options...

OptionsWrapper?  It won't be exposed really since users will almost always just use `Options.Create`?

HaoK: Or we actually could keep this class private too right?  The factory method is all they need

davidfowl: Will do since people may think Options is the non generic base of `Options<T>`. Keeping them separate seems like a good idea.

davidfowl: Keeping it private feels a bit extreme. That came to mind but feels pretty awful :smile:

davidfowl:
:up: :date:

HaoK:
A bit orthogonal to this, we should maybe rename the normal implementation class OptionsManager => SingletonOptions if we are going to have multiple IOptions implementations by default now.

`SingletonOptions<>`
`OptionsWrapper<>`
`OptionsMonitor<>`
`Options.Create<>`

davidfowl:
SharedOptions sounds good actually.

test/Microsoft.Extensions.Options.Test/OptionsTest.cs
```diff
@@ -302,5 +302,26 @@ public static TheoryData Configure_GetsEnumOptionsFromConfiguration_Data
...
+        [Fact]
+        public void OptionsWrapper_MakesOptions()
+        {
+            var options = new OptionsWrapper<FakeOptions>(new FakeOptions
+            {
+                Message = "This is a message"
+            });
+
+            Assert.Equal("This is a message", options.Value.Message);
+        }
```
HaoK: Should we add one test that does the override just to make sure the behavior is defined now?

```
services.AddOptions();
services.AddSingleton<IOptions<Foo>>(Options.Create<Foo>(myfoo)>);
```

HaoK:
Ok I can do the rename to SharedOptions in a seperate PR

test/Microsoft.Extensions.Options.Test/OptionsTest.cs
davidfowl: Will do

HaoK:
:shipit: looks good to me

lodejard:
Why does SharedOptions sound good?

I like three change otherwise, registering a closed generic was always intended as a way to "take control" if you have a special case.

davidfowl:
@lodejard is `ConfigurableOptions<T>` better?

HaoK:
I was just trying to suggest some names that give a bit more meaning to what OptionsManager does.

OptionsManager is a very generic name that doesn't convey any of the usage semantics, not as big a deal when it was the only IOptions, but now that we also have OptionsWrapper...

OptionsManager is really the shared/cached implementation of IOptions, while the wrapper is a stand alone instance of IOptions that you would use when you don't want to cache...

OptionsMonitor aka Reloadable options is kind of a third implementation but it doesn't implement IOptions yet (we have an issue tracking whether it makes sense to unify IOptionsMonitor/IOptions https://github.com/aspnet/Options/issues/100)

src/Microsoft.Extensions.Options/Options.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace Microsoft.Extensions.Options
```
JunTaoLuo: No copyright info?

PR #110, opened 2016-01-14
Remove ConfigureOptions sugar for types/instances

HaoK:
cc @divega

### DIFF
--- a/src/Microsoft.Extensions.Options/OptionsServiceCollectionExtensions.cs
+++ b/src/Microsoft.Extensions.Options/OptionsServiceCollectionExtensions.cs
@@ -24,72 +24,6 @@ namespace Microsoft.Extensions.DependencyInjection
             return services;
         }
 
-        private static bool IsAction(Type type)
-        {
-            return (type.GetTypeInfo().IsGenericType && type.GetGenericTypeDefinition() == typeof(Action<>));
-        }
-
-        private static IEnumerable<Type> FindIConfigureOptions(Type type)
-        {
-            var serviceTypes = type.GetTypeInfo().ImplementedInterfaces
-                .Where(t => t.GetTypeInfo().IsGenericType && t.GetGenericTypeDefinition() == typeof(IConfigureOptions<>));
-            if (!serviceTypes.Any())
-            {
-                var error = "No IConfigureOptions<> found.";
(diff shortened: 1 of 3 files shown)


src/Microsoft.Extensions.Options/OptionsServiceCollectionExtensions.cs
```diff
@@ -24,72 +24,6 @@ public static IServiceCollection AddOptions(this IServiceCollection services)
...
-
-            var serviceTypes = FindIConfigureOptions(configureInstance.GetType());
-            foreach (var serviceType in serviceTypes)
-            {
-                services.AddSingleton(serviceType, configureInstance);
-            }
-            return services;
-        }
-
         public static IServiceCollection Configure<TOptions>(this IServiceCollection services, Action<TOptions> setupAction)
```
divega: `setupAction` is a very weird parameter name given there nothing called setup around this code any more. `configureAction`?

HaoK: Don't we typically call this configureOptions in other places?

divega: Do we? I am not aware.

HaoK: Well actually AddIdentity/AddMvc call this setupAction still

divega: We shouldn't :disappointed:

HaoK: But I typically name `Action<TOptions>` configureOptions, these days anyways... AddAuthentication has it this way...

divega: I am ok with making it consistently `configureOptions` if that is your preference.

HaoK: Sure lets go with that

HaoK:
0002d4ebc03bdd44270714748a84625f16fc2ebe

PR #112, opened 2016-01-24
New Extension for easier configuration

stevenpwalsh05:
`services.AddOptions().Configure<Data>(Configuration.GetSection("Data"));`

Feels redundant to me, so this extension should allow configuration to look like this instead:

`services.AddOptions().Configure<Data>(Configuration);`

### DIFF
--- a/src/Microsoft.Extensions.Options.ConfigurationExtensions/OptionsServiceCollectionExtensions.cs
+++ b/src/Microsoft.Extensions.Options.ConfigurationExtensions/OptionsServiceCollectionExtensions.cs
@@ -26,6 +26,22 @@ namespace Microsoft.Extensions.DependencyInjection
             return services;
         }
 
+        public static IServiceCollection Configure<TOptions>(this IServiceCollection services, IConfigurationRoot root)
+            where TOptions : class
+        {
+            if (services == null)
+            {
+                throw new ArgumentNullException(nameof(services));
+            }
+
+            if (root == null)
+            {
+                throw new ArgumentNullException(nameof(root));
+            }
+            
+            return Configure<TOptions>(services, root.GetSection(typeof(TOptions).Name));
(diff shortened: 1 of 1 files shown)


davidfowl:
This is a pretty weird extension to add as it seems to only work for the scenario where the top level class name exactly matches a config section...

stevenpwalsh05:
You're right, this would not cover 100% of configuration use cases, however I assume it's the most "common" use case.  At least in my own experience it seems like what I do most often.  For my own efforts, this is how i'd prefer the API to look, but if you think it adds extra confusion to the API of course I understand.

davidfowl:
> You're right, this would not cover 100% of configuration use cases, however I assume it's the most "common" use case. At least in my own experience it seems like what I do most often. For my own efforts, this is how i'd prefer the API to look, but if you think it adds extra confusion to the API of course I understand.

I don't care about the 100% case, it's really 2 things:

1. Is it common enough the it warrants a first class API. How big is the cliff when you need to something slightly off the default path.
2. Is it obvious what the API is doing and how easy is it to turn off if it doesn't do what I expect.

cwe1ss:
I think that this method is not obvious. According to the name you wouldn't expect it to read the values from a section. It would need a different name - e.g. `ConfigureFromTypeNameSection<Data>(Configuration)` or `ConfigureByConvention<Data>(Configuration)`. However now it is almost the same amount of code as the current solution. For this reason, I don't think that it should be integrated.

divega:
I agreed with what @davidfowl said. We don't necessarily want a convention that matches the type name to one configuration key on the root to become a thing, at least not yet. It might be truth that without it there is a bit of extra code you need to write often but at least the source of the configuration data remains obvious and the APIs remain flexible. We can always revisit this in the future, but closing for now.

PR #114, opened 2016-02-11
Add TryConfigure method

henkmollema:
Adds a `TryConfigure<T>` method which configures the given configuration action using `TryAddSingleton` so it only can be configured once.

### DIFF
--- a/src/Microsoft.Extensions.Options/OptionsServiceCollectionExtensions.cs
+++ b/src/Microsoft.Extensions.Options/OptionsServiceCollectionExtensions.cs
@@ -2,9 +2,6 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
-using System.Collections.Generic;
-using System.Linq;
-using System.Reflection;
 using Microsoft.Extensions.DependencyInjection.Extensions;
 using Microsoft.Extensions.Options;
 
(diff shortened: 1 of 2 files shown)


herecydev:
What kind of use case do you see for this? One of the problems I foresee is that you don't know to what extend the first configure has configured. I.e. If you had 2 properties, how do you know which properties have been configured.

One of the fundamentals of the current structure is that you can incrementally add to the options as it's passed to each `IConfigureOptions<TOptions>`. This can be checked within each configure action (check for null) and you can prevent overwriting "already set" properties. I assume this was what you were looking for?

HaoK:
I think the usage for a TryConfigure would basically be only to supply a default override, basically only configure the options if noone else has...  I'm not convinced this is something we really need either, but I don't think its harmful to have...  @divega ?

divega:
Yeah, sounds like it would only be useful to provide a default if no one has, but generally you can do this by moving your call to `Configure()` up to be the first one, so unless you are a library that doesn't have access to change the code in startup you shouldn't ever really need this method.

@henkmollema what are we missing?

herecydev:
The only harm derives from the assumption that it is "fully configured" from the first time. Without peeking the properties you just don't know until the consumer deals with it. I'm not convinced and feel it could be misused for "try add *these* properties" but that's ultimately a learning exercise.

henkmollema:
Yes, what @HaoK says is my use case. I'm building a component-based system where each component can set defaults to options when the consumer doesn't configure them explicitly.

henkmollema:
I agree that it's a rather specific use-case and might be prone to misconfiguration. I'll leave it up to you to decide whether it's useful enough.

HaoK:
My feeling is that this probably isn't worth adding at this point, since this extension method can live in your system for now, if we get more requests for this, we can always add this later.  @divega thoughts?

divega:
Agreed so closing the PR for now. Thanks @henkmollema and @herecydev for the details.

PR #115, opened 2016-02-11
Remove unsused test class

henkmollema:
This file lives in `test/Microsoft.Extensions.OptionsModel.Test`. Probably missed out when committing the big rename.


HaoK:
Good catch thanks!  merged 02e8d7f9cd7100a8580e31817bdcbcf498e8cd7d

Issue #82, opened 2015-09-18
API Design: make strongly typed settings easier


jods4:
I am lifting some discussion that was briefly started by @brockallen in #31 into its own issue.

I have been using the `IOptions` API for a while now in the context of application settings and it just feels convoluted for no reason. I have the impression that I repeat meaningless, verbose, boilerplate code because... well because the framework wants me to.

Basically, anywhere a component needs to adjust its behaviour based on a setting (typically stored in `config.json`), you see a pattern similar to:
```csharp
class FrobController
{
  private readonly AppSettings settings;

  public FrobController(IOptions<AppSettings> options)
  {
    this.settings = options.Options; // or options.Value in beta-8 I think
  }
 
  public void FrobMe()
  {
    for (int i = 0; i < settings.FrobbingTimes; i++)
      FrobOnce();
  }
}
```

And if you use settings in many places or have split your settings in several classes by domain, you repeat that ctor code quite a lot. What bothers me?

- What the hell is this `IOptions<>`? Vague naming is vague. And it's not the _options_ since they are in `AppSettings`, actually.
- What is it about? Why should I use it? What does it buy me? In this use-case: nothing!
- Given its generic name and its non-existent added value, it's hard to discover. Trust me, the first time you encounter this you try to DI `AppSettings`. And it fails. Then you Google.
- The dance around `.Options` or `.Value` is superfluous.
- The naming is awkward. What should I call the parameter `IOptions`? Any name I could come up with is either long for what it's worth (`optionsAccessor`) or misleading/clashing with the _actual_ options object (`AppSettings`).

In #31 @lodejard said:
> it is very similar, but the accessor service has advantages in that the creation of the singleton is deferred to when it's first used.

Deferred creation is something that needs a general solution, not a domain-specific one. And the solution is injecting `Lazy<T>` and registering a factory for `T`. And given the smallish nature of application settings in general and the fact that they are guaranteed to be used, the lazy initialization is probably more overhead than benefit.

> Plus it enables the pattern where multiple IConfigureOptions can be added as services to alter any given TOptions. 

Maybe for middleware configs or other advanced stuff. I don't see how that would be a use case for application settings. The use-case here is really to parse config into a strongly-typed object and make it available to the whole app.

> With a singleton MyOptions added to DI it must be fully baked before it is added, and the things doing the options configuration all need to do their work before IOC is available.

[...]

Issue #31, opened 2014-12-13
Rename IOptions<TOptions>.Options to .Value


davidfowl:
Normally, you declare variables like this:

```C#
public class Foo
{
     private MyOptions _options;
     public Foo(IOptions<MyOptions> options)
     {
        _options = options.Value;
     }
}
```

HaoK:
@davidfowl @lodejard @divega with Named options being removed (soon), should we go back to IOptionsAccessor to match IHttpContextAccessor?

Or keep it IOptions and just switch it to IOptions.Value?

HaoK:
Options are meant to be simple poco data objects. I believe the motivations are similar to why configuration is not put into DI. But the actual initialization of options(IConfigureOptions) are.

brockallen:
But `IOptions<T>`, in effect, is putting options into DI. I guess I don't understand the difference.

HaoK:
@divega @davidfowl What do you guys think about renaming IOptions to ISharedOptions to convey that this is meant to be used mostly when options are singletons/shared... 

ISharedOptions.Value would be its only property.

lodejard:
Lets keep the name IOptions<T>. It started out as IOptionsAccessor which was inconveniently verbose.

@brockallen it is very similar, but the accessor service has advantages in that the creation of the singleton is deferred to when it's first used. Plus it enables the pattern where multiple IConfigureOptions<TOptions> can be added as services to alter any given TOptions. With a singleton MyOptions added to DI it must be fully baked before it is added, and the things doing the options configuration all need to do their work before IOC is available.

HaoK:
c623fbc3c6a8b037ed191da950587d6fa884bcac

PR #116, opened 2016-02-15
IOptions not required in classes using POCO options class

matthewDDennis:
# Problem
Currently, using the` IOptions<TOptions>` pattern requires that the class using the `TOptions` have a dependency on` IOptions<TOptions>` rather than `TOptions`. This forces the class to have a dependency on `Microsoft.Extensions.Options`.  This is not desirable for classes that could be used outside of ASPNET Core classes or apps.

Ideally, the class should only be dependent on the `TOptions` POCO class.

# Solution
By adding a 
``` c#
services.AddSingleton<TOptions>(sp => sp.GetRequiredService<IOptions<TOptions>>().Value);
```
After each
``` c#
services.AddSingleton<IConfigureOptions<TOptions>>(new ConfigureOptions<TOptions>(configureOptions));
```
This allows the `ServiceProvider` to resolve `TOptions` to a correctly initialize instance of `TOption` when used in an ASPNET Core app, and to an manually initialized instance in cases where the ASPNET Core Dependency Injection and Configuration is not used.
`

### DIFF
--- a/src/Microsoft.Extensions.Options.ConfigurationExtensions/OptionsServiceCollectionExtensions.cs
+++ b/src/Microsoft.Extensions.Options.ConfigurationExtensions/OptionsServiceCollectionExtensions.cs
@@ -10,7 +10,7 @@ namespace Microsoft.Extensions.DependencyInjection
     public static class OptionsConfigurationServiceCollectionExtensions
     {
         public static IServiceCollection Configure<TOptions>(this IServiceCollection services, IConfiguration config)
-            where TOptions : class
+            where TOptions : class, new()
         {
             if (services == null)
             {
@@ -23,11 +23,12 @@ namespace Microsoft.Extensions.DependencyInjection
             }
 
             services.AddSingleton<IConfigureOptions<TOptions>>(new ConfigureFromConfigurationOptions<TOptions>(config));
+            services.AddSingleton<TOptions>(sp => sp.GetRequiredService<IOptions<TOptions>>().Value);
             return services;
         }
 
         public static IServiceCollection Configure<TOptions>(this IServiceCollection services, IConfiguration config, bool trackConfigChanges)
-            where TOptions : class
+            where TOptions : class, new()
         {
             if (services == null)
             {
@@ -40,6 +41,7 @@ namespace Microsoft.Extensions.DependencyInjection
             }
 
             services.AddSingleton<IConfigureOptions<TOptions>>(new ConfigureFromConfigurationOptions<TOptions>(config));
+            services.AddSingleton<TOptions>(sp => sp.GetRequiredService<IOptions<TOptions>>().Value);
             if (trackConfigChanges)
             {
                 services.AddSingleton<IOptionsChangeTokenSource<TOptions>>(new ConfigurationChangeTokenSource<TOptions>(config));
--- a/src/Microsoft.Extensions.Options.ConfigurationExtensions/OptionsServiceCollectionExtensions.cs
+++ b/src/Microsoft.Extensions.Options.ConfigurationExtensions/OptionsServiceCollectionExtensions.cs
@@ -4,6 +4,7 @@
 using System;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.Options;
+using Microsoft.Extensions.DependencyInjection.Extensions;
 
 namespace Microsoft.Extensions.DependencyInjection
 {
@@ -23,7 +24,7 @@ namespace Microsoft.Extensions.DependencyInjection
             }
 
             services.AddSingleton<IConfigureOptions<TOptions>>(new ConfigureFromConfigurationOptions<TOptions>(config));
-            services.AddSingleton<TOptions>(sp => sp.GetRequiredService<IOptions<TOptions>>().Value);
+            services.TryAddSingleton<TOptions>(sp => sp.GetRequiredService<IOptions<TOptions>>().Value);
             return services;
         }
 
@@ -41,7 +42,7 @@ namespace Microsoft.Extensions.DependencyInjection
             }
 
             services.AddSingleton<IConfigureOptions<TOptions>>(new ConfigureFromConfigurationOptions<TOptions>(config));
-            services.AddSingleton<TOptions>(sp => sp.GetRequiredService<IOptions<TOptions>>().Value);
+            services.TryAddSingleton<TOptions>(sp => sp.GetRequiredService<IOptions<TOptions>>().Value);
             if (trackConfigChanges)
             {
                 services.AddSingleton<IOptionsChangeTokenSource<TOptions>>(new ConfigurationChangeTokenSource<TOptions>(config));
--- a/src/Microsoft.Extensions.Options/OptionsServiceCollectionExtensions.cs
+++ b/src/Microsoft.Extensions.Options/OptionsServiceCollectionExtensions.cs
@@ -25,7 +25,7 @@ namespace Microsoft.Extensions.DependencyInjection
         }
 
         public static IServiceCollection Configure<TOptions>(this IServiceCollection services, Action<TOptions> configureOptions)
-            where TOptions : class
+            where TOptions : class, new()
         {
             if (services == null)
             {
@@ -38,6 +38,7 @@ namespace Microsoft.Extensions.DependencyInjection
             }
 
             services.AddSingleton<IConfigureOptions<TOptions>>(new ConfigureOptions<TOptions>(configureOptions));
+            services.AddSingleton<TOptions>(sp => sp.GetRequiredService<IOptions<TOptions>>().Value);
             return services;
         }
     }
--- a/src/Microsoft.Extensions.Options/OptionsServiceCollectionExtensions.cs
+++ b/src/Microsoft.Extensions.Options/OptionsServiceCollectionExtensions.cs
@@ -38,7 +38,7 @@ namespace Microsoft.Extensions.DependencyInjection
             }
 
             services.AddSingleton<IConfigureOptions<TOptions>>(new ConfigureOptions<TOptions>(configureOptions));
-            services.AddSingleton<TOptions>(sp => sp.GetRequiredService<IOptions<TOptions>>().Value);
+            services.TryAddSingleton<TOptions>(sp => sp.GetRequiredService<IOptions<TOptions>>().Value);
             return services;
         }
     }
--- a/test/Microsoft.Extensions.Options.Test/OptionsTest.cs
+++ b/test/Microsoft.Extensions.Options.Test/OptionsTest.cs
@@ -339,5 +339,26 @@ namespace Microsoft.Extensions.Options.Tests
             var sp = services.BuildServiceProvider();
             Assert.Equal("Override", sp.GetRequiredService<IOptions<FakeOptions>>().Value.Message);
         }
+
+        [Fact]
+        public void Options_CanGetOptionsFromServiceProvider()
+        {
+            // copy  an existing SetupCallsInOrder test to set things up
+            var services = new ServiceCollection().AddOptions();
+            var dic = new Dictionary<string, string>
+            {
+                {"Message", "!"},
+            };
+            var builder = new ConfigurationBuilder().AddInMemoryCollection(dic);
+            var config = builder.Build();
+            services.Configure<FakeOptions>(o => o.Message += "Igetstomped");
+            services.Configure<FakeOptions>(config);
+            services.Configure<FakeOptions>(o => o.Message += "a");
+            services.Configure<FakeOptions>(o => o.Message += "z");
+
+            var options = services.BuildServiceProvider().GetService<FakeOptions>();
+            Assert.NotNull(options);
+            Assert.Equal("!az", options.Message);
+        }
     }
 }


davidfowl:
The thing you mention as the problem is part of the design. If you don't want to depend on options then don't use this project. There's a standalone configuration binder that can give you what you need. You can also avoid the dance around registering a callback and register the poco object however you like. 

The intent of IOptions<T> is to have some indirection so the pattern of IConfigureOptions<T> can be used to compose different component setting options.

For that reason I'd reject this change as it goes against the intended design.

matthewDDennis:
This proposal does nothing to change the use of `IConfigureOptions` to compose options.
All it does is prevent the need to pass in an `IOptions<TOptions>` in the constructor of the consuming class, only to reference the value property.

By registering a service for `TOptions` as `sp => sp.GetRequiredService<IOptions<TOptions>>().Value` all the goodness of the current functionality for configuring the options values is still used.  When the `TOption` is resolved from the `ServiceProvider`, the `IOptions<TOption>` implementation `OptionsManager` is still created with the `IEnumerable<Action<TOption>> `collection of register setups.

All my simple change does is
- Eliminate the need to pass in an` IOptions<TOptions>` in the consuming class' constructor, only to get the Value.  This is accomplished by moving the accessing of the value from a `IOptions<TOptions>` passed to the class, to the DI container with an `ImplementationFactory `for the `TOption` which just gets the value from the` OptionsManager `(the `IOptions<>`  implementation)
- Eliminate the dependency of the consuming class on the `IOptions` interface and all its dependencies.

The consuming class doesn't need or care about `IOptions`, only the `TOptions` value.  Passing in an `IOptions` is leaking implementation details that the consuming class should not need to know about.

However, if you have a current implementation the uses `IOptions<TOption>` as the constructor dependency, the code will still work, as this change does not change any of the existing service registrations associated with the `Options`, only adds one more registration that allows the constructor dependency to be an instance `TOptions` class, without the need for accessing the Value property.

davidfowl:
I understand the change I just don't agree with it nor do I think it should be in the box. 

> Eliminate the need to pass in anIOptions<TOptions> in the consuming class' constructor, only to get the Value. This is accomplished by moving the accessing of the value from a IOptions<TOptions> passed to the class, to the DI container with an ImplementationFactoryfor the TOption which just gets the value from theOptionsManager(the IOptions<> implementation)

> Eliminate the dependency of the consuming class on the IOptions interface and all its dependencies.

I think that is a non goal of this project. If you want that then like I said, add the factory yourself or add the poco directly. `IOptions<T>` isn't meant to replace that.

matthewDDennis:
Your call, but thinking it over, I realize that the `IOptions` interface is only required in the internal implementation to provide a means to allow multiple steps of configuration to be execute the **first **time the TOptions value is requested.  There is no benefit to the consumer of the TOption to be passed an IOptions<TOption> at all.  It only increases the complexity (very slightly to be sure), and adds an unnecessary dependency which is not small.  In most cases, an class of options should only be dependent on `System` as the properties are only int, bool, string, TimeSpans,  or other simple types.

Don't get me wrong, I think the Options design and implementation is amazing, and makes configuring a system extremely flexible and easy.  It just seems like you ran 99 yards, and stopped at the 1 yard line.

matthewDDennis:
Others have also tried to explain this issue. See issues #82 and #31.

HaoK:
Now that there is nothing useful on the IOptions interface (RIP NamedOptions) other than the Value, I agree that it would be nicer to just take the Options class directly in ctors.  Even if we are just saving everyone from calling IOptions.Value, that seems worth considering if there's no loss in flexibility or functionality...

This also would have a nice benefit of letting us remove the options dependencies from several of our repos.

divega:
I was thinking in the same lines as @haok. With no named options in the picture and given the ability to register a `Func<IServiceProvider,TService>` in DI, I cannot think of a reason doing what is proposed in this PR would be wrong. 

`IOptions<T>` still provides additional features, e.g. the lazyness (nothing happens until the first access to `.Value`) and potentially change notification if we end up merging that in, but for the most common usage depending on just the `TOptions` is rather nice.

The only reason I can speculate we wouldn't want this is if we decided that consuming options needs to be a bit more cumbersome than consuming any standard class just to make it clear at consuming time that they get created differently... But personally I see the seemlessnes of this proposal as a nice feature.

davidfowl:
> I was thinking in the same lines as @haok. With no named options in the picture and given the ability to register a Func<IServiceProvider,TService> in DI, I cannot think of a reason doing what is proposed in this PR would be wrong.

3 against 1! That's fine. I'd caution against using delegates by default, it's less performant than binding against types directly.

> IOptions<T> still provides additional features, e.g. the lazyness (nothing happens until the first access to .Value) and potentially change notification if we end up merging that in, but for the most common usage depending on just the TOptions is rather nice.

So what do we recommend to users and what do we do with our middleware? (for example). Writing code is easy, looking at the impact is what I'm having a hard time with. Do we tell people to call `Configure<T>(...)` and then look for a `T` in their classes? What happens when we have change notification, is there a grow up story where you have to use `IOptions<T>` or is it basically just a lazy `T`?

lodejard:
Yeah, using the TOptions as a service instead of IOptions<TOptions> does break the graph each place it occurs. Even if it was available I think we'd still keep the IOptions interface in our own code for perf reasons...

divega:
Yeah, I think it would be something like a grow up story, e.g. we would tell customers to only use `IOptions<T>` if you need the extra stuff it provides.

BTW, I don't think the lazyness part has a big impact. I think it would go from "nothing happens until someone uses the value" to "nothing happens until someone resolves an instance of `TOptions`". Whether this was intended or not in most of our current usages we access the value in the constructor of the class that depends on `IOptions<T>`

Let me also clarify that for the time being change notification on options requires you to depend on yet another interface, but I see an opportunity to converge on a single interface (see #100).

HaoK:
I think if we take this change, this would be one big reason why we shouldn't collapse change notifications into IOptions.  

It seems a bit weird that we would add this feature but not use it ourselves, are we sure the perf difference is large enough to not have one consistent usage pattern for options (IOptions would basically be buried if we switched to this pattern)

divega:
Sure we can keep using `IOptions<T>` in our own code. To @haok's point on the advantage of not having to depends on options, if we want it to be an even lighter weight dependency to have, we should consider moving the interfaces to an abstractions package.

divega:
@haok not sure I understand what you mean but we can chat offline.

davidfowl:
> It seems a bit weird that we would add this feature but not use it ourselves, are we sure the perf difference is large enough to not have one consistent usage pattern for options (IOptions would basically be buried if we switched to this pattern)

We should obviously measure but the delegate pattern is less efficient in general. You're basically hiding the dependency graph from the DI system. I'm worried about the concept count as well. You need to be able to describe to people why `IOptions<T>` exists and when it should be used.

> Sure we can keep using IOptions<T> in our own code. To @haok's point on the advantage of not having to depends on options, if we want it to be an even lighter weight dependency to have, we should consider moving the interfaces to an abstractions package.

Is options heavyweight?

/cc @DamianEdwards

HaoK:
Options is pretty light right now, it depends on only DI.Abstractions and System.Primitives (which is from the change notifications, so we could split that one out potentially)

HaoK:
Assuming we are able to switch to TOptions usage everywhere, the answer to when you should use IOptions is never right?  It becomes purely an implementation mechanism for how we were able to get all of the IConfigureOptions stuff to work...

davidfowl:
> Assuming we are able to switch to TOptions usage everywhere, the answer to when you should use IOptions is never right? It becomes purely an implementation mechanism for how we were able to get all of the IConfigureOptions stuff to work...

I don't see the point in keeping `IOptions<T>` if we go with this approach (assuming the performance doesn't suffer).

divega:
We need something we can call in the delegate :smile:

HaoK:
Well I guess we can remove it entirely if we want to explicitly depend on our implementation of `OptionsManager<T>` I guess...

davidfowl:
Or we can keep it and say use `IOptions<T>` if you want the performance benefit :smile:

lodejard:
I dunno... Hiding the call to `IOptions<T>.Value`... Breaks the service graph and obfuscates where the T is coming from... The T is a model rather than a service after all... I got a really bad feeling about this...

After all we're in the middle of trying to remove GetService calls and factory delegates right now, aren't we?

davidfowl:
> I dunno... Hiding the call to IOptions<T>.Value... Breaks the service graph and obfuscates where the T is coming from... The T is a model rather than a service after all... I got a really bad feeling about this...

That's my gut as well. It's like making removing `IHttpContextAccessor` and instead using a delegate to represent injecting the `HttpContext` directly, it smells bad.

> After all we're in the middle of trying to remove GetService calls and factory delegates right now, aren't we?

Yap, because performance.

Can we agree that this should be an addon?

HaoK:
That is a good point about `IHttpContextAccessor`, as it seems like both of these should have the same behavior.  The code still read betters with direct injection since you can't do anything else with `IHttpContextAccessor/IOptions`...

Maybe just inject extension methods instead of directly doing it as part of Configure?
```
  services.InjectHttpContext();
  services.InjectOptions<TOptions>();
```

lodejard:
I'm still on the side of the fence that we don't do either - I think the point behind hiding the http context access or was that it is also a bad idea.

For example, if we hide the IOptions for end-users it'll look like the T is a service, and if they register the T class like they do any other service it'll work - but only for half of the consumers.

Imho, I don't think any good can come from wallpapering over how things really work with an illusion. After all, you are in reality still getting an IOptions.Value - if there's a real problem with that it's one thing - but I'm still uncomfortable with a purely cosmetic change that obfuscates and degrades the system...

matthewDDennis:
I don't think there is a performance issue at all.  The delegate is registered as a Singleton, so it will be evaluated on first use and cached in the `ServiceProvider`.  It should actually be faster as there will no longer be a call to `OptionsManager.Value`.

I believe the DI code is compiling the delegates, so they should be fast.  If they aren't they should.  When writing Munq DI, I found this to be a major performance boost.  Basically, do the expensive stuff once and early, on registration if possible, and on first use if not.

HaoK:
re @lodejard 

Isn't the main point of this change (if we make it opt in via a new Inject like method) to give developers the option to ask for this cosmetic behavior if they desire.  This is just sugar for making hiding these easier, unless you are saying that people should not be encouraged to hide this at all?

davidfowl:
I'm with @lodejard on this change. It's something that can be trivially added to any change and I don't think we should do it by default.

> I don't think there is a performance issue at all. The delegate is registered as a Singleton, so it will be evaluated on first use and cached in the ServiceProvider. It should actually be faster as there will no longer be a call to OptionsManager.Value.

Not quite, I don't have any numbers but creating a type is always faster than using a delegate registration.

lodejard:
Yeah, actually another issue could be if nobody calls Configure for a BlahOptions then an `IOptions<BlahOptions>.Value` dependency will return a default new BlahOptions(), but a `BlahOptions` dependency will be a missing service...

Urrr.... That said... It's true if `IOptions<TOptions>.Value` is for the immutable case always... At the risk of opening a whole new can of worms... We could consider changing the pattern to consider the TOptions class itself the service, and move away from the `IOptions<TOptions>.Value` altogether in favor of an options constructor pattern.

```cs
public class BlahOptions
{
    public BlahOptions(IEnumerable<IConfigureOptions<BlahOptions>> alterations = null) 
    {
        if (alterations != null) foreach(var alteration in alterations)
        {
            alteration.Configure(this);
        }
    }
}
```

That way depending on BlahOptions as a service would actually be the correct thing to do all the time `public class Blah { public Blah(BlahOptions options) {...} ... }`. Down-side is that BlahOptions takes a boiler-plate constructor, the options POCO depends on IConfigureOptions interface, and must be registered as an `.AddSingleton<BlahOptions>()` service, but on the plus side there's no need for an `OptionsManager<TOptions>` or `IOptions<TOptions>` constructor dependency and it's slightly less magical how the system works.

HaoK:
I sort of like it, but I'd prefer combining/hiding the registration and boilerplate into a new `services.AddOptions<T>` which could setup a delegate to invoke the IConfigureOptions and register the result as a singleton all at once.  Its a bit more magical, but enables the POCOs to remain clean like today...

lodejard:
Well, the whole idea is to avoid the using the **last resort** features like delegate services and IServiceProvider.GetService. Those techniques should be used when there's literally no other way to accomplish a dependency in a declarative way - not as an alternative to having your actual dependencies declared as constructor arguments...

HaoK:
In that case, this has nudged me back towards `IOptions.Value` as the lesser of the two evils... 

I really like being able to use any vanilla POCO Options class, and the fact that the new requirements for making options a service are in your face, it feels a lot more complicated...

matthewDDennis:
Due to your concern about performance I decided to measure getting the `IOptions<TOptions>` and the `TOptions` from a `ServiceProvide`r.  Measured each in a loop of 100000.  I used code from one of the tests.  Contrary to your expectations, the `TOptions` version takes half the time as the `IOptions<TOptions>` version.

Interestingly, the value returned is incorrect.  It appears that the
```c#
            services.Configure<FakeOptions>(config);
```
is not working.

I'm running RC1-update.

## Results
IOptions<FakeOptions>.Value.Message has incorrect value of "Igetstompedaz"
FakeOptions.Message has incorrect value of "Igetstompedaz"
IOptions<FakeOptions> took 00:00:00.0292005
FakeOptions took 00:00:00.0150615
Do it again in reverse order
FakeOptions took 00:00:00.0150976
IOptions<FakeOptions> took 00:00:00.0276157
Press ENTER to exit
## Code

```c#
using System;
using System.Collections.Generic;
using System.Diagnostics;
using System.Linq;
using System.Threading.Tasks;
using Microsoft.Extensions.Configuration;
using Microsoft.Extensions.DependencyInjection;
using Microsoft.Extensions.OptionsModel;

namespace OptionsTest
{
    public class Program
    {
        public static void Main(string[] args)
        {
            const int NumLoops = 100000;
            Stopwatch stopwatch = new Stopwatch();

            // Setup the DI
            var services = new ServiceCollection().AddOptions();
            var dic = new Dictionary<string, string>
            {
                {"Message", "!"},
            };
            var builder = new ConfigurationBuilder().AddInMemoryCollection(dic);
            var config = builder.Build();
            services.Configure<FakeOptions>(o => o.Message += "Igetstomped");
            services.Configure<FakeOptions>(config);
            services.Configure<FakeOptions>(o => o.Message += "a");
            services.Configure<FakeOptions>(o => o.Message += "z");
            services.AddSingleton<FakeOptions>(sp => sp.GetRequiredService<IOptions<FakeOptions>>().Value);

            // get the FakeOptions to ensure the ServiceProvider has resolved and cached the results.
            var serviceProvider = services.BuildServiceProvider();
            var iopFake1 = serviceProvider.GetRequiredService<IOptions<FakeOptions>>();
            if (iopFake1.Value.Message != "!az")
                Console.WriteLine($"IOptions<FakeOptions>.Value.Message has incorrect value of \"{iopFake1.Value.Message}\"");
            var fake1 = serviceProvider.GetRequiredService<FakeOptions>();
            if (fake1.Message != "!az")
                Console.WriteLine($"FakeOptions.Message has incorrect value of \"{fake1.Message}\"");

            string message;
            stopwatch.Restart();
            for (int i = 0; i < NumLoops; i++)
                message = serviceProvider.GetRequiredService<IOptions<FakeOptions>>().Value.Message;
            var IOpTime = stopwatch.Elapsed;

            stopwatch.Restart();
            for (int i = 0; i < NumLoops; i++)
                message = serviceProvider.GetRequiredService<FakeOptions>().Message;
            var POCOTime = stopwatch.Elapsed;

            // and do it again in reverse order
            stopwatch.Stop();
            Console.WriteLine($"IOptions<FakeOptions> took {IOpTime}");
            Console.WriteLine($"FakeOptions took {POCOTime}");
            Console.WriteLine("Do it again in reverse order");

            stopwatch.Restart();
            for (int i = 0; i < NumLoops; i++)
                message = serviceProvider.GetRequiredService<FakeOptions>().Message;
            POCOTime = stopwatch.Elapsed;

            stopwatch.Restart();
            for (int i = 0; i < NumLoops; i++)
                message = serviceProvider.GetRequiredService<IOptions<FakeOptions>>().Value.Message;
            IOpTime = stopwatch.Elapsed;

            // and do it again in reverse order
            stopwatch.Stop();
            Console.WriteLine($"FakeOptions took {POCOTime}");
            Console.WriteLine($"IOptions<FakeOptions> took {IOpTime}");
            Console.WriteLine("Press ENTER to exit");

            Console.ReadLine();

        }
    }
}
```

HaoK:
Nice, I was just about to do this myself :)  So the delegate usage is faster than IOptions (which is using an open generic service)

divega:
@HaoK if you haven't already, could you try this micro-benchmark with our current bits to double check? 

Anyway, I suspect the perf concern with delegate registration applies more to larger graphs of dependencies, e.g. activate a controller that depends on a few services them themselves depend on options. I also suspect that @matthewDDennis is onto something when he says that the service being a singleton will make a difference. 

> Interestingly, the value returned is incorrect.

Sounds like we could have a bug :disappointed:. @HaoK can you please take a look?

HaoK:
@matthewDDennis I was going to run something like this, feel free to rerun your benchmarks with these scenarios that don't have any of the test code that might muddle things up. @divega I can take a look, its likely some weirdness with the test though as this particular test is doing a bunch of weird stuff.

Scenario A: IOptions today unchanged with controller
```
public class OptionsA { 
  public string Stuff { get; set; }
}
public class ControllerA {
    private OptionsA _o;
    public ControllerA(IOptions<OptionsA> options) { _o = options.Value; }
}

public void ScenarioA() {
 services.AddOptions();
 services.Configure<OptionsA>(o => o.Stuff = "Yes");
 services.AddTransient<ControllerA>();
 // measure this
 var controller = sp.GetService<Controller>();
}
```

Scenario B: Using this PR's behavior
```
public class OptionsB { 
  public string Stuff { get; set; }
}
public class ControllerB {
    private OptionsB _o;
    public ControllerB(OptionsB options) { _o = options; }
}

public void ScenarioB() {
 services.AddOptions();
  services.Configure<OptionsB>(o => o.Stuff = "Yes");
services.AddTransient<ControllerB>();
 // measure this
 var controller = sp.GetService<ControllerB>();
}
```

Scenario C: Using this @lodejard 's pattern
```
public class OptionsC {
   public OptionsC(IEnumerable<IConfigureOptions<OptionsC> configs) {
       foreach (var c in configs) c.Configure(this);
   }

  public string Stuff { get; set; }
}
public class ControllerC {
    private OptionsC _o;
    public ControllerC(OptionsC options) { _o = options; }
}

public void ScenarioC() {
 services.AddTransient<ControllerC>();
 services.Configure<OptionsC>(o => o.Stuff = "Yes");
 services.AddSingleton<OptionsC>();
 // measure this
 var controller = sp.GetService<ControllerC>();
}
```

Scenario D: My hybrid approach
```
public class OptionsD { 
  public string Stuff { get; set; }
}
public class ControllerD {
    private OptionsD _o;
    public ControllerD(OptionsD options) { _o = options; }
}

public void ScenarioD() {
 services.AddTransient<ControllerD>();
 services.Configure<OptionsD>(o => o.Stuff = "Yes");
 services.AddSingleton<OptionsD>(sp => {
    var o = new OptionsD();
    foreach (var c in sp.GetRequiredService<IEnumerable<IConfigureOption<OptionD>>>())
        c.Configure(o);
    return o;
 });
 // measure this
 var controller = sp.GetService<ControllerD>();
}
```

matthewDDennis:
It looks like the error in the resolve value in my perf program may have been due to the version.  I pull down the sources and ran against them and got the correct value.

HaoK:
I ran some numbers with these 4 scenarios with a very basic controller/setup.  

The obvious difference between today and all 3 options is the switch away from IOptions.Value, I'm guessing the large hit is purely based on the OptionsManager's EnsureLazyInitialized call in IOptions.Value.

It does confirm that there is indeed a (slight) perf difference as we expected from each pattern.

For Iterations = 50000000;

Perf Today: 18.8780305s
PerfPR: 12.1215447s
PerfLou: 11.9652762s
PerfHao: 12.377871s

Full test code below:

```C#
        private const int Iterations = 50000000;

        public class ControllerNow {
            public OptionsNow O;
            public ControllerNow(IOptions<OptionsNow> options) { O = options.Value; }
        }
        public class OptionsNow { public string Stuff { get; set; } }

        [Fact]
        public void PerfNow()
        {
            services.AddOptions();
            services.AddTransient<ControllerNow>();
            services.Configure<OptionsNow>(o => o.Stuff = "Whatever");

            // prime it once
            Assert.Equal("Whatever", sp.GetRequiredService<ControllerNow>().O.Stuff);
            var stopwatch = new Stopwatch();
            stopwatch.Start();
            for (var i=0; i< Iterations; i++) Assert.Equal("Whatever", sp.GetRequiredService<ControllerNow>().O.Stuff);
            stopwatch.Stop();
            Assert.Equal(0, stopwatch.Elapsed.TotalSeconds);
        }

        public class ControllerPR {
            public OptionsPR O;
            public string Stuff;
            public ControllerPR(OptionsPR options) { O = options; }
        }
        public class OptionsPR { public string Stuff { get; set; } }

        [Fact]
        public void PerfPR()
        {
            services.AddOptions();
            services.AddTransient<ControllerPR>();
            services.AddSingleton(s => s.GetRequiredService<IOptions<OptionsPR>>().Value);
            services.Configure<OptionsPR>(o => o.Stuff = "Whatever");

            // prime
            Assert.Equal("Whatever", sp.GetRequiredService<ControllerPR>().O.Stuff);
            var stopwatch = new Stopwatch();
            stopwatch.Start();
            for (var i = 0; i < Iterations; i++) Assert.Equal("Whatever", sp.GetRequiredService<ControllerPR>().O.Stuff);
            stopwatch.Stop();
            Assert.Equal(0, stopwatch.Elapsed.TotalSeconds);
        }

        public class ControllerLou {
            public OptionsLou O;
            public string Stuff;
            public ControllerLou(OptionsLou options) { O = options; }
        }
        public class OptionsLou {
            public OptionsLou(IEnumerable<IConfigureOptions<OptionsLou>> cs) { foreach (var c in cs) c.Configure(this); }
            public string Stuff { get; set; }
        }

        [Fact]
        public void PerfLou()
        {
            services.AddTransient<ControllerLou>();
            services.AddSingleton<OptionsLou>();
            services.Configure<OptionsLou>(o => o.Stuff = "Whatever");

            // prime
            Assert.Equal("Whatever", sp.GetRequiredService<ControllerLou>().O.Stuff);
            var stopwatch = new Stopwatch();
            stopwatch.Start();
            for (var i = 0; i < Iterations; i++) Assert.Equal("Whatever", sp.GetRequiredService<ControllerLou>().O.Stuff);
            stopwatch.Stop();
            Assert.Equal(0, stopwatch.Elapsed.TotalSeconds);
        }

        public class ControllerHao {
            public OptionsHao O;
            public ControllerHao(OptionsHao options) { O = options; }
        }
        public class OptionsHao { public string Stuff { get; set; } }

        [Fact]
        public void PerfHao()
        {
            services.AddTransient<ControllerHao>();
            services.AddSingleton(s => {
                var o = new OptionsHao();
                foreach (var c in s.GetRequiredService<IEnumerable<IConfigureOptions<OptionsHao>>>())
                    c.Configure(o);
                return o;
            });
            services.Configure<OptionsHao>(o => o.Stuff = "Whatever");

            // prime
            Assert.Equal("Whatever", sp.GetRequiredService<ControllerHao>().O.Stuff);
            var stopwatch = new Stopwatch();
            stopwatch.Start();
            for (var i = 0; i < Iterations; i++) Assert.Equal("Whatever", sp.GetRequiredService<ControllerHao>().O.Stuff);
            stopwatch.Stop();
            Assert.Equal(0, stopwatch.Elapsed.TotalSeconds);
        }
```

cwe1ss:
I agree that injecting `SomeOptions` directly would be nice. This would remove the need to reference any Options-package for most libraries. However I also agree that injecting regular objects is a little bit weird. 

Would it help somehow to move the dependency from the service to the Options POCO? `SomeOptions` could have to implement a marker interface (ideally `IOptions` but this would be a big breaking change). The DI process and ConfigureOptions() could now verify that only objects implementing this interface are used for this. But since this still requires a dependency on the Options-package, I don't think that this would be an improvement.

This is also related to the question I just asked: [Why no Microsoft.Extensions.Options.Abstractions?](https://github.com/aspnet/Options/issues/120) (sorry for not reading this PR first). I think the package structure is not consistent with other Extensions-projects. In case we have to keep the `IOptions<TOptions>` usage, it would feel better to me to have a dependency on an `Abstractions` package with no other dependencies at all. Most libraries don't care about the actual implementation of the Options system. This way it would be similar to injecting an ILoggerFactory and so on. 
This would also remove the need to have the `Microsoft.Extensions.Options.ConfigurationExtensions`-package. Instead, we would just have `Microsoft.Extensions.Options` and `Microsoft.Extensions.Options.Abstractions`.

matthewDDennis:
@HaoK Your tests are using AddTransient instead of AddSingleton for the TOption.  This will make a difference in your measurements.  I think Singleton is the more likely pattern to use here, unless you are working with auto reloading on config file changes.

matthewDDennis:
BTW, I just did a performance comparison of Munq DI with the ASPNET Core DI.  Surprisingly, the ASPNET DI came out slightly faster, and has better error checking for cyclic dependencies. I am impressed.

Not sure why someone would want to use Autofac or Structuremap instead based on performance numbers.  Granted, there are a few features they have that aren't in the ASPNET DI, but I really don't think that they are important in a properly designed system.

HaoK:
@matthewDDennis I wanted the controllers to be transient, the options are registered as singletons.  If both were registered as singletons, I would hope all of these tests end up close to identical times after this many iterations...

HaoK:
One other nice benefit of treating Options as a service would be letting us normalize `Configure<T>` into => `AddOptions<T>` like the rest of our frameworks...

```C#
 // Just registers the MyOptions singleton
services.AddOptions<MyOptions>()

// Register the options singleton and configure it
services.AddOptions<MyOptions>(options => options.Setting = "Whatever"); //
```

HaoK:
Thanks for this PR, we discussed this today and are going in a similar direction with the following changes outlined here: https://github.com/aspnet/Options/issues/122

Issue #122, opened 2016-02-26
Removing IOptions<T>


HaoK:
Summary of planned options changes:

1. Removing `IOptions<T>` entirely, Options must be registered as services (updating all middleware/frameworks)
2. `AddOptions<T>` sugar which registers an options as a singleton:

```
            services.AddSingleton(s => {
                var o = new OptionsHao();
                foreach (var c in s.GetRequiredService<IEnumerable<IConfigureOptions<OptionsHao>>>())
                    c.Configure(o);
                return o;
            });
```
3.  Our framework options will not use this sugar, but will switch to ctor's taking `IEnumerable<IConfigureOptions<T>>` (with potentially a new base Options<T> class)

cc @divega @lodejard @davidfowl

leastprivilege:
finally.

davidfowl:
One thing I don't want to lose: The ability for middleware get a default `TOptions` when nothing is specified. That is one advantage of having a default `IOptions<TOptions>` available in the container.

HaoK:
Hrm, that's definitely an issue right now.  

Does making options the last parameter and optional work?

```C#
  public Middleware(Foo f, MyOptions options = null) {
    _options = options ?? new MyOptions();
  }
```

PinpointTownes:
> One thing I don't want to lose: The ability for middleware get a default TOptions when nothing is specified. That is one advantage of having a default IOptions<TOptions> available in the container.

:+1:

matthewDDennis:
I think; this and discussions in #116 show that the DI system is missing an important feature.  Many DI containers are able to resolve concrete classes, even if they have not been registered in the container.  Having done this in Munq, and having a quick look at the DI code, it shouldn't be a difficult task to add this feature.  All the hard parts are already done. Just need to create a factory method for the class if it does not exist in the resolved classes, and register it.

By default, the class would be resolved Transient, but could be explicitly registered for other lifetimes.

I'll take a look and see if I can put together a pull request for the DI container.

davidfowl:
> Many DI containers are able to resolve concrete classes, even if they have not been registered in the container.

How many? The top 10 (if there are that many)? That's not something we want to do by default in all cases anyways. In the cases where we need control over what things come from DI and what things don't, we manually type activate.

> Having done this in Munq, and having a quick look at the DI code, it shouldn't be a difficult task to add this feature.

[...]

PR #123, opened 2016-02-26
RIP IOptions

HaoK:
Part 1 of https://github.com/aspnet/Options/issues/122

- Removes IOptions, AddOptions, OptionsWrapper
+ Adds AddOptions<T>, AddOptionsMonitor<T>

cc @divega @lodejard @davidfowl


src/Microsoft.Extensions.Options/OptionsServiceCollectionExtensions.cs
```diff
@@ -12,18 +10,37 @@ namespace Microsoft.Extensions.DependencyInjection
...
 
+
+        public static IServiceCollection AddOptions<TOptions>(this IServiceCollection services) where TOptions : class, new()
+        {
+            if (services == null)
+            {
+                throw new ArgumentNullException(nameof(services));
+            }
+
+            services.AddSingleton(serviceProvider =>
```
divega: Should we make this `TryAdd` so that `AddOptions` is idempotent? (you could argue it would need to be called `TryAddOptions` then, but I think we have done this before without renaming the method).

HaoK: I went back and forth on this one myself, if @lodejard wants this to be more of a service as opposed to a framework, I think this should not be a TryAdd...

lodejard: Well guessed. It could also be `AddSingletonOptions<TOptions>` and `TryAddSingletonOptions<TOptions>` to be closer to the service methods they correspond to. Though I'm guessing you wouldn't prefer that. :)

halter73: Who call's AddOptions these days? Is it always done directly by the application?

Does this mean that any service that takes options is going to need an optionless ctor?

HaoK: None of our repos are going to call the `AddOptions<T>` sugar by default, instead they will register the singleton in our AddXyz methods.  Apps or other frameworks are free to use the sugar method to avoid requiring complicated ctors for their options.  

Services can certainly adopt the 2 ctors pattern as well.  But I don't think we need to do that in general since most of our frameworks fall over anyway if AddXyz isn't called.

divega:
It might not be relevant to the PR,  but `AddTransientOptions` hasn't been discussed :smile:

HaoK:
Replaced by https://github.com/aspnet/Options/pull/124

PR #124, opened 2016-03-02
Options Reborn

HaoK:
- New API portion of https://github.com/aspnet/Options/issues/122, removal of IOptions will come later
- IConfigureOptions => IOptionsConfiguration
- IOptionsInitializer service instead of IEnumerableConfigureOptionsMegaLongTypeName
- Hybrid where old ConfigureOptions implements both new and old to keep existing code working

Example:
```
        public class EnumOptions
             public EnumOptions() { }
             public EnumOptions(IOptionsInitializer<EnumOptions> initializer) : this()
             {
                 initializer.Initialize(this);
             }
```
cc @divega @davidfowl @lodejard

### DIFF
--- a/src/Microsoft.Extensions.Options.ConfigurationExtensions/ConfigureFromConfigurationOptions.cs
+++ b/src/Microsoft.Extensions.Options.ConfigurationExtensions/ConfigureFromConfigurationOptions.cs
@@ -6,7 +6,7 @@ using Microsoft.Extensions.Configuration;
 
 namespace Microsoft.Extensions.Options
 {
-    public class ConfigureFromConfigurationOptions<TOptions> : ConfigureOptions<TOptions>
+    public class ConfigureFromConfigurationOptions<TOptions> : ConfigureOptions<TOptions>, IOptionsConfiguration<TOptions>
         where TOptions : class
     {
         public ConfigureFromConfigurationOptions(IConfiguration config)
--- a/src/Microsoft.Extensions.Options.ConfigurationExtensions/OptionsServiceCollectionExtensions.cs
+++ b/src/Microsoft.Extensions.Options.ConfigurationExtensions/OptionsServiceCollectionExtensions.cs
@@ -23,6 +23,7 @@ namespace Microsoft.Extensions.DependencyInjection
             }
 
             services.AddSingleton<IConfigureOptions<TOptions>>(new ConfigureFromConfigurationOptions<TOptions>(config));
+            services.AddSingleton<IOptionsConfiguration<TOptions>>(new ConfigureFromConfigurationOptions<TOptions>(config));
             return services;
         }
 
--- a/src/Microsoft.Extensions.Options/IOptionsConfiguration.cs
+++ b/src/Microsoft.Extensions.Options/IOptionsConfiguration.cs
@@ -1,0 +1,10 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.Extensions.Options
+{
+    public interface IOptionsConfiguration<in TOptions> where TOptions : class
+    {
+        void Configure(TOptions options);
+    }
+}
--- a/src/Microsoft.Extensions.Options/IOptionsInitializer.cs
+++ b/src/Microsoft.Extensions.Options/IOptionsInitializer.cs
@@ -1,0 +1,15 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.Extensions.Options
+{
+    public interface IOptionsInitializer<TOptions> where TOptions : class
+    {
+        /// <summary>
+        /// Initializes the options instance.
+        /// </summary>
+        /// <param name="instance">The options instance.</param>
+        /// <returns></returns>
+        TOptions Initialize(TOptions instance);
+    }
+}
--- a/src/Microsoft.Extensions.Options/OptionsConfiguration.cs
+++ b/src/Microsoft.Extensions.Options/OptionsConfiguration.cs
@@ -5,9 +5,9 @@ using System;
 
 namespace Microsoft.Extensions.Options
 {
-    public class ConfigureOptions<TOptions> : IConfigureOptions<TOptions>, IOptionsConfiguration<TOptions> where TOptions : class
+    public class OptionsConfiguration<TOptions> : IOptionsConfiguration<TOptions> where TOptions : class
     {
-        public ConfigureOptions(Action<TOptions> action)
+        public OptionsConfiguration(Action<TOptions> action)
         {
             if (action == null)
             {
--- a/src/Microsoft.Extensions.Options/OptionsInitializer.cs
+++ b/src/Microsoft.Extensions.Options/OptionsInitializer.cs
@@ -1,0 +1,36 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.Extensions.Options
+{
+    public class OptionsInitializer<TOptions> : IOptionsInitializer<TOptions> where TOptions : class
+    {
+        private readonly IEnumerable<IConfigureOptions<TOptions>> _configurations;
+
+        public OptionsInitializer(IEnumerable<IConfigureOptions<TOptions>> configurations)
+        {
+            _configurations = configurations;
+        } 
+
+        /// <summary>
+        /// Initializes the options instance.
+        /// </summary>
+        /// <param name="instance">The options instance.</param>
+        /// <returns></returns>
+        public virtual TOptions Initialize(TOptions instance)
+        {
+            if (instance == null)
+            {
+                throw new ArgumentNullException(nameof(instance));
+            }
+            foreach (var configure in _configurations)
+            {
+                configure.Configure(instance);
+            }
+            return instance;
+        }
+    }
+}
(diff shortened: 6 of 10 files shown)


HaoK:
FYI @divega brought up a good point when we discussed this today, that the error story for this sucks...  If you forget to include a ctor that takes the initializer...  Your configures just get ignored...silently :/

lodejard:
That is true of the entire change to make options a service, not just the initializer

HaoK:
That is true of the entire change to make options a service, not just the initializer

Well if you forget to add the options class you get a fairly obvious service type missing exception. If you forget to have the constructor nothing fails. You just get unfortunate behavior of always having default options (except when using reload able options which will do always do the right thing and apply your configurations for you :)

HaoK:
This is a benefit of the conflated AddOptions of T alternative as it's not possible to forget one of the steps and end up in this weird half state.

src/Microsoft.Extensions.Options.ConfigurationExtensions/ConfigureFromConfigurationOptions.cs
```diff
@@ -6,7 +6,7 @@
 
 namespace Microsoft.Extensions.Options
 {
-    public class ConfigureFromConfigurationOptions<TOptions> : ConfigureOptions<TOptions>
+    public class ConfigureFromConfigurationOptions<TOptions> : ConfigureOptions<TOptions>, IOptionsConfiguration<TOptions>
```
divega: I like in principle the renaming of the interface very much. Using a verb in `IConfigureOptions` made it hard to name things that interacted with it or that implemented it. 

I wonder if this goes far enough though. The main remaining problem I believe is that we are overloading the word Configuration to mean `Microsoft.Extensions.Configuration` and also "the thing that you do to mutate Options". 

That makes it almost impossible to find a good name for this class. I.e. `ConfigurationOptionsConfiguration` would probably be the name we would come up with if we just applied the standard rules to name this class: it is just an `OptionsConfiguration` that is based on `Configuration`.

I wonder if we can come up with a better word for "the thing that you do to mutate options" and apply it to the interface name and the service collection extension, e.g.:
- `IOptionsSetup` the original name we had wasn't that bad IMO
- `IOptionsMutator` it is probably the most mathematically correct but feels too obscure.

Any thoughts on this?

HaoK: Lou had proposed Alteration, I suggested your OptionsConfiguration since I remembered you liked that previously.  I always liked Setup personally…

divega: - `IOptionsConfigurationAction` doesn't actually help much with naming this class but maybe it makes more sense in a way.

divega: Alternation seems mathematically correct and a bit less obscure that mutator or mutation, at least IMO :smile:

Yeah, I like setup too.

divega: Just throwing more words at it: 

- modifier
- modification

src/Microsoft.Extensions.Options.ConfigurationExtensions/OptionsServiceCollectionExtensions.cs
```diff
@@ -23,6 +23,7 @@ public static IServiceCollection Configure<TOptions>(this IServiceCollection ser
             }
```
divega: The name of this class should mean "here are the extensions for adding option "configurations" that are based on values from `Microsoft.Extensions.Configuration`. If we go with `IOptionsConfiguration` for the interface, then the current name is a bit ambiguous. It should probably be revisited no matter what we do.

HaoK: The name of this class can be as ugly as you want, noone uses it, its a static placeholder class for extensions... so go nuts :)

divega: I will try to make it pretty anyway :smile:

src/Microsoft.Extensions.Options.ConfigurationExtensions/ConfigureFromConfigurationOptions.cs
HaoK: I'll switch IOptionsConfiguration => IOptionsSetup in the next iteration (assuming we move forward)

src/Microsoft.Extensions.Options/ConfigureOptions.cs
```diff
@@ -5,7 +5,7 @@
 
 namespace Microsoft.Extensions.Options
 {
-    public class ConfigureOptions<TOptions> : IConfigureOptions<TOptions> where TOptions : class
+    public class ConfigureOptions<TOptions> : IConfigureOptions<TOptions>, IOptionsConfiguration<TOptions> where TOptions : class
```
divega: We should revisit the name of this class based on what we decide for the interface.

src/Microsoft.Extensions.Options/OptionsConfiguration.cs
```diff
@@ -0,0 +1,32 @@
...
+        {
+            if (action == null)
+            {
+                throw new ArgumentNullException(nameof(action));
+            }
+
+            Action = action;
+        }
+
+        public Action<TOptions> Action { get; private set; }
```
davidfowl: C# 6 brah

```C#
public Action<TOptions> Action { get;  } 
```

davidfowl:
> FYI @divega brought up a good point when we discussed this today, that the error story for this sucks... If you forget to include a ctor that takes the initializer... Your configures just get ignored...silently :/

I was going to bring this up today... It sucks hard. We might have to tell people to use the callback thingy.

src/Microsoft.Extensions.Options/IOptionsConfiguration.cs
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.Extensions.Options
+{
+    public interface IOptionsConfiguration<in TOptions> where TOptions : class
+    {
+        void Configure(TOptions options);
```
divega: We should consider revisiting the name of this method if we agree that we don't want to overload "configure" and "configuration".

divega: I get that it may be a pretty big change.

HaoK:
@davidfowl Well my personal preference is to dispense with the OptionsInitializer/ctor, and just require `AddOptions<T>` since today an `AddSingleton<T>` is required anyways..... kill two birds with one stone, no dangling wtf situations...

src/Microsoft.Extensions.Options/IOptionsConfiguration.cs
HaoK: Shrug, we can just go from Configure => Setup not a huge deal

src/Microsoft.Extensions.Options/IOptionsInitializer.cs
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.Extensions.Options
+{
+    public interface IOptionsInitializer<TOptions> where TOptions : class
```
divega: :+1: Even if I still don't know how we can round up the whole story we tell customers about how to define their own options, I think this is awesome.

src/Microsoft.Extensions.Options/IOptionsInitializer.cs
```diff
@@ -0,0 +1,15 @@
...
+    {
+        /// <summary>
+        /// Initializes the options instance.
+        /// </summary>
+        /// <param name="instance">The options instance.</param>
+        /// <returns></returns>
+        TOptions Initialize(TOptions instance);
```
divega: Nit: The `TOptions` parameter in `IOptionsConfiguration.Configure()` is called `options`. I prefer `instance` and I think we should make it consistent.

divega:
> I was going to bring this up today... It sucks hard. We might have to tell people to use the callback thingy.

Agreed. It seems a bit unprecedented to have a story that customers can get wrong so easily. I am open to accept that this is just a services thing and that we need to get used to it, but still.

divega:
By the way, it is possible to have a base class for options that does the trick and it doesn't look bad on usage:

``` C#
    public class OptionsBase<TOptions> where TOptions: OptionsBase<TOptions>
    {

        public OptionsBase(IOptionsInitializer<TOptions> initializer)
        {
            initializer.Initialize((TOptions)this);
        }
    }

    public class MyOption : OptionsBase<MyOption>
    {
        public MyOption(IOptionsInitializer<MyOption> initializer) : base(initializer)
        {
        }

        public string Name { get; set; }
    }
```

HaoK:
We decided against this change

PinpointTownes:
> We decided against this change

Any chance you could share more details? :smile: 
I was not a huge fan of the options as services approach, but this PR came with great naming changes.

HaoK:
Which naming changes in particular did you like?  We couldn't really come to a consensus that the new approach was better than the current.

PinpointTownes:
> Which naming changes in particular did you like?

`IConfigureOptions` -> `IOptionsConfiguration` or `IOptionsSetup`
`ConfigureOptions` -> `OptionsConfiguration` or `OptionsSetup`

divega:
I have created a bug to follow up with some possible renames: https://github.com/aspnet/Options/issues/127.

PR #125, opened 2016-03-03
Target minimal TFMs

pranavkm:


pranavkm:
cc @davidfowl

test/Microsoft.Extensions.Options.Test/project.json
```diff
@@ -17,15 +18,12 @@
...
-        "Microsoft.NETCore.Platforms": "1.0.1-*",
         "System.Dynamic.Runtime": "4.0.11-*"
       }
     },
     "net451": {
-      "frameworkAssemblies": {
-        "System.Runtime": ""
-      },
       "dependencies": {
+        "System.Runtime": "4.1.0-*",
```
davidfowl: wut?

pranavkm: It's the same issue as Configuration. Guard rail fails saying it can't find a runtime lib for `System.Runtime`.

davidfowl: Is that because net core platforms is being used for all frameworks?

PR #130, opened 2016-03-11
Target dnx451

bricelam:
Reverts ee8b434bf268ba297e2e9c360af45bab018f1f37 for DNX


divega:
:shipit:

bricelam:
Recalling. After chatting with @Eilon, we decided that adding the facades to user's `project.json` is an acceptable workaround.

PR #132, opened 2016-03-25
Options: Move interfaces into Abstractions

HaoK:
Did we also want to combine Options + ConfigurationExtensions now that we have an abstractions?

@divega


Eilon:
I suppose there will be follow-up PRs to this to simplify dependencies in other repos? I.e. refer to abstractions instead of options?

divega:
> follow-up PRs to this to simplify dependencies in other repos?

@eilon I was talking to @HaoK about this earlier. We believe that should be possible in some cases (e.g. mudflewares) but we are not sure about the size of the set of repositories that consume options but don't use the functionality in the Options package to add them to the service container.  If you think we should prioritize this for RC2,  we look into it.

divega:
> Did we also want to combine Options + ConfigurationExtensions

I don't think we should do that if it means that a lot of packages will start depending transitively from Configuration that don't do it today. So I guess it depends on the answer to @Eilon's question.

Eilon:
Hmm I guess I'm curious then - if we don't think our own libraries can use just the abstractions, does that diminish the value of this change? (I'm a big fan of abstractions in general, but sometimes abstractions just aren't enough.)

HaoK:
Well if we also move the DI extensions into abstractions that would be sufficient for most use cases.

divega:
> if we also move the DI extensions into abstractions that would be sufficient... 

But wouldn't that require moving the implementation types registered by the extension methods into the abstraction package as well? If yes, how much is left out?

HaoK:
Basically only the handful of OptionsMonitor implementation classes.

divega:
Just OptionsMonitor? Wouldn't it include OptionsManager as well? Anyway, it seems to me that such split would be very arbitrary and that we need to decide if we take this PR as is based on how much value we see for libraries (ours or third party) which only need to consume options or provide their own implementations of the interfaces. And I suspect there is little value.

HaoK:
Actually we might be able to move options manager now that hosting is the one that calls AddOptions instead of each repo...  But I am happy to leave things the way they are since it doesn't appear that we are getting much value of adding Abstractions, Options is basically pretty close to what we would want Abstractions to be already I guess.

davidfowl:
:disappointed: why?

HaoK:
Do you disagree with any of the discussion above for why we didn't think the abstractions was going to be super useful?

Issue #150, opened 2016-06-07
Don't rely on IoC containters to preserve registration order IConfigureOptions services


halter73:
Autofac and possibly other IoC containers will not return IEnumerable resolved services in the order they were registered, but Options currently [depends on this behavior](https://github.com/aspnet/Options/blob/c7bff7184d3a3ab4074859b3fd3d36d83c474357/test/Microsoft.Extensions.Options.Test/OptionsTest.cs#L143). This causes certain extension methods in Kestrel, WebListener, MVC (and EF?) to break when Autofac is used.

This will probably require adding an `Order` property back to `IConfigureOptions` similar to what we had [back in beta7](https://github.com/aspnet/Options/blob/1.0.0-beta7/src/Microsoft.Framework.OptionsModel/IConfigureOptions.cs).

A key difference is we want to continue running the actions passed to [Configure](https://github.com/aspnet/Options/blob/4ef5a92054013f528d54ee3270501743f8d7e255/src/Microsoft.Extensions.Options/OptionsServiceCollectionExtensions.cs#L38) in the order they are added. This will likely require that `Configure` inspect the `IServiceCollection` prior to registering the `ConfigureOptions` instance.

This should resolve the issues raised in https://github.com/aspnet/Hosting/issues/785

divega:
A few more details:
1. We should bring back the overloads of `Configure<T>()` that allow setting the order. 
2. When `Configure<T>()` inspects existing instances of `IConfigureOptions<T>` we should do (pseudo code) `1 + (setups.Any() ? setups.Max(s => s.Order) : 0)`. This works better for custom order than other options.
3. This will be a breaking change for any custom implementation of `IConfigureOptions<T>`, including our own in  https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNetCore.Mvc.Core/Internal/MvcCoreMvcOptionsSetup.cs and https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNetCore.Mvc.WebApiCompatShim/WebApiCompatShimOptionsSetup.cs.

HaoK:
We decided we are going to require IoC containters to preserve registration order

PR #151, opened 2016-06-07
Restore Order to the World!

HaoK:
Fixes https://github.com/aspnet/Options/issues/150

Currently is doing a basic auto-incrementing order logic instead of inspecting the service collection for each call to Configure. This means Configures without order specified will be clustered together.  

This lets users have a bit more granular control if they want to put overrides with large orders, and preserves call ordering when order is never specified.

### DIFF
--- a/src/Microsoft.Extensions.Options.ConfigurationExtensions/OptionsConfigurationServiceCollectionExtensions.cs
+++ b/src/Microsoft.Extensions.Options.ConfigurationExtensions/OptionsConfigurationServiceCollectionExtensions.cs
@@ -22,6 +22,21 @@ namespace Microsoft.Extensions.DependencyInjection
         public static IServiceCollection Configure<TOptions>(this IServiceCollection services, IConfiguration config)
             where TOptions : class
         {
+            return services.Configure<TOptions>(config, Options.Options.NextDefaultConfigureOptionsOrder() + Options.Options.DefaultConfigurationBindOrderOffset);
+        }
+
+        /// <summary>
+        /// Registers a configuration instance which TOptions will bind against.
+        /// </summary>
+        /// <typeparam name="TOptions">The type of options being configured.</typeparam>
+        /// <param name="services">The <see cref="IServiceCollection"/> to add the services to.</param>
+        /// <param name="config">The configuration being bound.</param>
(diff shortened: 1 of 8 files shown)


HaoK:
@divega @davidfowl @Eilon @halter73 @Tratcher @muratg

HaoK:
Note the order is increment for all calls to Configure where order is not specified irrespective of TOptions, since I didn't think the actual values matter much so long as they are always increasing and with large enough gaps between for hardcoded values to interleave.

src/Microsoft.Extensions.Options/Options.cs
```diff
@@ -6,6 +6,30 @@
     public static class Options
     {
         /// <summary>
+        /// The order which will be used for the next call to Configure
+        /// </summary>
+        public static int DefaultConfigureOptionsOrder;
```
divega: Statics + the ability to specify the order explicitly can results into non-deterministic behavior if you create more than one container in the same process or AppDomain.

HaoK:
Fair enough, updated default order logic to the following:

Defaults to 1 (non config binder), -10000 for config binder
Otherwise uses 10+Max of any IConfigureOption of the right type that has an ImplementationInstance

src/Microsoft.Extensions.Options.ConfigurationExtensions/OptionsConfigurationServiceCollectionExtensions.cs
```diff
@@ -22,6 +23,31 @@ public static class OptionsConfigurationServiceCollectionExtensions
         public static IServiceCollection Configure<TOptions>(this IServiceCollection services, IConfiguration config)
             where TOptions : class
         {
+            var setups = services.Where(s => s.ServiceType == typeof(IConfigureOptions<TOptions>) && s.ImplementationInstance != null && s.ImplementationType == typeof(ConfigureFromConfigurationOptions<TOptions>));
+            var order = -10000; // REVIEW: should we expose this default constant for config?
```
divega: I don't think we should go back to special casing the order of configuration binding at all. Simple order of registration provides enough control and enough rules for anyone to understand :smile:. We have lived without it since beta 8 and personally I haven't seen feedback indicating that it is missed.

HaoK:
Closing in favor of requiring 3rd party containers to respect registration order

PR #152, opened 2016-08-10
Add an overload for Configure where the action can accept an IServiceProvider

tuespetre:


HaoK:
Is the main motivation of this change to be able to consume services in your Configure?  If so, you can just implement `IConfigureOptions<YourOptions>` explicitly and add it to DI, with the services required taken as ctor arguments.

HaoK:
In general we are trying to avoid service locator, so we are not keen on adding overloads that take IServiceProvider

tuespetre:
There is a line between "service locator is bad" and "sweet Jesus, we have to write *more code* just to satisfy some ideal of what is and is not an anti-pattern", and I feel very confident in saying that I have found that line. 

It's alright, I'll just add this to my "Extensions.cs" file that I copy-paste-replace in every project. ;)

davidfowl:
@tuespetre do you end up with the bulk of your config in the lambda? Can you show where you use this today?

tuespetre:
@davidfowl well, that depends -- when 'the bulk of the config' amounts to setting a ton of properties on an `Options`-y class, then the answer could be yes. But if there is some complex construction, for instance, classes that derive from `JsonInputFormatter` and `JsonOutputFormatter` and need `ObjectPoolProvider` and `ArrayPool<char>` and a whole bunch of other stuff, that will go in an `IConfigureOptions<MvcOptions>` , and `services.Configure<MvcOptions>(mvc => ...)` will just be used for the rest.

It looks like *every* time I've wanted this (plus the one time I've seen someone else want it), it's been for the purpose of copying things from one `Options`-y class to another `Options`-y class, and I've just needed **one** thing from the services. I don't see it as a super common thing to do though, so I'm okay with keeping it in my 'bag of extensions.' We can directly use `services.Configure<IConfigureOptions<MvcOptions>>(provider => new ConfigureOptions<MvcOptions>(options => { ... }));` anyways, but I thought the shorthand would be nicer to have around. :smile_cat:

HaoK:
Revisiting due to https://github.com/aspnet/Options/pull/239

HaoK:
Going to hand merge this as a new method on IOptionsBuilder instead of on IServiceCollection as an extension method.

HaoK:
Will be added to: https://github.com/aspnet/Options/pull/239

HaoK:
Actually doesn't appear that we really need this, since this is a specialized version of `Configure<TDep>` where TDep is IServiceProvider, the only slight difference is `Configure<TDep>` is registered as a transient instead of singleton, but in reality that won't matter since its the IOptionsFactory's lifetime that really matters (since that hangs onto the list of IConfigureOptions anyways).

But I will add a test to IOptionsBuilder to verify that this is possible, i.e.:
```
            var someService = new SomeService("Something");
            var services = new ServiceCollection().AddOptions().AddSingleton(someService);
            services.AddOptions<FakeOptions>().Configure<IServiceProvider>((o, s) => o.Message = s.GetRequiredService<SomeService>().Stuff);

            var factory = sp.GetRequiredService<IOptionsFactory<FakeOptions>>();
            Assert.Equal("Something", factory.Create(Options.DefaultName).Message);

```

tuespetre:
@HaoK that's a neat solution! 👏

Issue #145, opened 2016-05-26
Revisit OptionsMonitor/ReloadableOptions/ChangeTracking


HaoK:

arielcloud:
is there a deadline for this feature?
is there a way for now to reload options on configuration change?

divega:
Assigning to 1.1.0 after talking to @glennc about @DamianEdwards feedback on this.

MisterJames:
@HaoK  I see that this is still needing design, but if this is something that can be worked on by the community I'd like to take a stab at this (or part of it).

HaoK:
`[IUpdateableOptions].Value` (name tweaking TBD)
Scoped cache of `IOptionsMonitor.CurrentValue`

divega:
We also decided that serviceCollection.Configure() extension method

HaoK:
How about `IOptionsState` instead of `IUpdateableOptions` giving us:

IOptions [Singleton permanent value]
IOptionsMonitor [Singleton permanent until notifications to trigger rebuild]
IOptionsState [ Captures the options state from the monitor for the scope of the request]

cc @davidfowl @DamianEdwards @Eilon

divega:
I prefer IOptionsSnapshot instead of IUpdatableOptions or IOptionsState.

cc @davidfowl @DamianEdwards @Eilon

HaoK:
Snapshot works for me as well

DamianEdwards:
Not feeling it. Need to stew on it some more...

Eilon:
Boo.

divega:
I think it us a pretty good name but besides,  `IOptionsSnapshot` is the first name we have discussed that accurately reflects what it is and doesn't convey wrong expectations, e.g. you would not expect to be able to load or update the options value using an instance of this service. 

Any problem if we go with it for an initial PR? We can keep looking and rename if we find something better.

Eilon:
Ok I admit it's actually growing on me (but that might just be the 🍺 talking). But yes a PR would be good anyway. Have plenty of time to rename if needed.

glennc:
This is a ping to make everyone think about the name again. All the ideas I have had are technically incorrect for one reason or other. So I am hoping someone throws out something amazing... :)

matthewDDennis:
Looking at what it does, I almost feels like an OptionsCache in that it holds a value that can be invalidated  base on a watcher or trigger. There isn't a timeout, but that would be easy to implement.

[...]

PR #154, opened 2016-09-28
Options Reloaded

HaoK:
Fixes https://github.com/aspnet/Options/issues/145#issuecomment-250020098

Initial shape with basic tests (want to add a few more)

Few points to consider:
- The feature is always 'on', any time Configure(IConfiguration) is called, we hook up the notifications (no way to disable)
- We could have the current OptionsManager implement both IOptions and IOptionsSnapshot (we'd register the class twice, singleton for IOptions, and Scoped for Snapshot, felt too weird so I didn't go that route, and copied the class, since its tiny)
- Support multiple listeners in OptionsMonitor.OnChange or not?  Basically subsequent calls could either add additional callbacks, or override the first callback.  I did the more complicated list of callbacks, but I'd be glad to kill this.
- (Found a nasty bug with CurrentValue in the old implementation where it wouldn't get updated if no one ever called OnChange)

### DIFF
--- a/src/Microsoft.Extensions.Options.ConfigurationExtensions/ConfigurationChangeTokenSource.cs
+++ b/src/Microsoft.Extensions.Options.ConfigurationExtensions/ConfigurationChangeTokenSource.cs
@@ -1,0 +1,40 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.Primitives;
+
+namespace Microsoft.Extensions.Options
+{
+    /// <summary>
+    /// Creates IChangeTokens so that IOptionsMonitor gets notified when IConfiguration changes.
+    /// </summary>
+    /// <typeparam name="TOptions"></typeparam>
+    public class ConfigurationChangeTokenSource<TOptions> : IOptionsChangeTokenSource<TOptions>
+    {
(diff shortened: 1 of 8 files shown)


HaoK:
@divega @Eilon @DamianEdwards @davidfowl

divega:
> The feature is always 'on', any time Configure(IConfiguration) is called, we hook up the notifications (no way to disable)

At first glance it looks like the baseline cost (if noone hooks up to listen) is a singleton IOptionsChangeTokenSource. Are there other hidden costs?

> Support multiple listeners in OptionsMonitor.OnChange or not? Basically subsequent calls could either add additional callbacks, or override the first callback. I did the more complicated list of callbacks, but I'd be glad to kill this.

I agree the code is a bit more involved this way, but I think it is fine, and I cannot see us getting away without it.

HaoK:
Yeah I guess the overhead isn't too bad if no one asks for an IOptionsMonitor since the change token hookup in the ctor won't happen

HaoK:
@Eilon @DamianEdwards @davidfowl last call for other name suggestions for the initial checkin...?  

'

galich:
Keep IOptionsSnapshot - it conveys the nature of the object ;)

HaoK:
acecacff57f9399e5871216a3f5a073548700dc3

RehanSaeed:
I fully expected the reason for the existence of `IOptions<>` being the reloading ability. Why was IOptionsSnapshot<> created instead and why not just register the options type  `T` directly?

Issue #161, opened 2017-01-18
Options 2.0 changes (named/validation/caching) for Auth 2.0


HaoK:
See if there's a way to reintroduce named options in a nicer way, such that we can consume it the same way today via IOptions/ISnapshotOptions.Value, where a new `IOptionsNameSelector` service would be the customization point where name(aka tenant) can be selected.

High level usage:
```C#
    // Add some kind of NamedInstance string property to IConfigure<TOptions>

    // Configure
    services.Configure<MyOptions>("Tenant1", o => o.Connection = "conn1");
    services.Configure<MyOptions>("Tenant2", o => o.Connection = "conn2");
    services.AddScoped<IOptionsNameSelector, MyTenantSelector>();
    
    public MyTenantSelector : IOptionsNameSelector {
        public MyTenantSelector(IHttpContextAccessor accessor);
        public string ResolveOptionsName() {
             return GetTenantNameFromRequest(accessor);
        }
    }

    public Controller(IOptions<MyOptions> options) {
       var tenantAwareOptions = options.Value;
    }

    public Controller(ISnapshotOptions<MyOptions> options) {
       var tenantAwareOptions = options.Value;
    }

```

The options implementations would use the `IOptionsNameSelector` and `NamedInstance` to return the appropriate options instance

Hopefully we can do this simply by enhancing the existing implementations of the current interfaces...  But its possible we might need to introduce `INamedOptions<MyOptions>`.

This would dovetail nicely with the auth changes that could take advantage of this feature via named auth scheme options

cc @divega @glennc

HaoK:
@divega Looks like you might get that `IOptionsCache` that you were hoping for.  Since `IOptionsNameSelector` potentially needs to be scoped, we can no longer register any of our options managers as singletons, and instead they will need to be scoped as well, which means we will need to push the cache instances into a shared singleton service instead to survive across requests since the lifetime of `IOptions<T>, IOptionsMonitor<T>` will now be scoped (but the actual options instances will continue to be stable, unless they are reloaded)

divega:
@HaoK cool :smile:

HaoK:
Looks like we will need to have more advanced cache capabilities with eviction to support multi-tenancy, we should be able to leverage the memory cache.

HaoK:
@divega updated the PR adding support for named options to `IOptionsMonitor`.  I also added a direct GetNamedInstance back to IOptions for a direct way to get a sibling.  

[...]

PR #162, opened 2017-01-19
Prototype named options

HaoK:
Exploring https://github.com/aspnet/Options/issues/161

Initial checkpoint adds namedInstance to IOptions (IOptionsSnapshot will be done in the next iteration)

Added a test that shows a simple scoped name selector that affects which options you get based on the scope

@divega @glennc


HaoK:
FYI this cache thing is a bit weird, since today IOptions and IOptionsMonitor don't share a cache, so we might have to have IOptionsCache and IOptionsMonitorCache [eww]

HaoK:
Replacing with new iteration

Issue #160, opened 2017-01-18
IOptionsSnapshot should support scoped IConfigureOptions (ie. DbContext)


HaoK:
Consuming a DbContext to configure options per request should be possible via.

```
   public class DbConfigure : IConfigureServices<MyOptions> {
       public DbConfigure(DbContext context) { Context = context; }
       public Configure(MyOptions options) {
           UseContextToConfigure(options);
       }
   }

   services.AddScoped<IConfigureService<MyOptions>, DbConfigure>();
```

This doesn't work today because we are caching the options until a change evicts (which works for the singleton IOptions but blocks this scenario)

cc @divega @glennc

HaoK:
c68ede33e50c585f71f4c86c9eff765b96d9acf4

PR #164, opened 2017-01-31
OptionsSnapshot always created per scope

HaoK:
Fixes https://github.com/aspnet/Options/issues/160

Unblocks using IConfigureOptions which have scoped dependencies like db context

cc @divega @glennc


src/Microsoft.Extensions.Options/OptionsSnapshot.cs
```diff
@@ -1,34 +1,21 @@
...
-        public virtual TOptions Value
-        {
-            get
-            {
-                return _options;
-            }
-        }
+        /// <param name="setups">The configuration actions to run.</param>
+        public OptionsSnapshot(IEnumerable<IConfigureOptions<TOptions>> setups) : base(setups)
+        { }
```
HaoK: Slight behavior difference with this change, previously we would eager cache .Value in the ctor, now we are lazy and don't cache until the first time .Value is used...  We should be intentional with how this works I guess...

HaoK: discussed with @divega and we will just have OptionsManager implement both interfaces

HaoK:
c68ede33e50c585f71f4c86c9eff765b96d9acf4

PR #165, opened 2017-02-01
Upgrade to VS 2017

natemcmaster:
Another vanilla upgrade.


src/Microsoft.Extensions.Options/Microsoft.Extensions.Options.csproj
```diff
@@ -0,0 +1,22 @@
...
+    <PackageTags>aspnetcore;options</PackageTags>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="1.2.0-*" />
+    <PackageReference Include="Microsoft.Extensions.Primitives" Version="1.2.0-*" />
+  </ItemGroup>
+
+  <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard1.0' ">
+    <PackageReference Include="System.ComponentModel" Version="4.4.0-*" />
```
BrennanConroy: netstandard1.0 is the only target framework, any reason not to include it with the other references

BrennanConroy:
🐑 :it:

src/Microsoft.Extensions.Options/Microsoft.Extensions.Options.csproj
natemcmaster: good catch. Code-gen strikes again 😈

PR #171, opened 2017-02-20
Add possibility to configure options validation in order to validate options at the start up time (not at runtime).

MaKCbIMKo:
This PR introduces the possibility to 'initialize' options at the start up time. It will prevent possible config issues that might occur in runtime because of incorrect config file.

Also, it allows to use initializers to write some custom logic to instantiate options objects.


src/Microsoft.Extensions.Options.Validation/OptionsValidationServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,84 @@
...
+            {
+                throw new ArgumentNullException(nameof(services));
+            }
+
+            services.AddSingleton<IOptionsValidatorManager, OptionsValidatorManager>();
+
+            return services;
+        }
+
+        public static IServiceCollection Validate<TOptions>(this IServiceCollection services, Expression<Func<TOptions, bool>> validateExpression)
```
HaoK: What do you think about a less restrictive signature for the validation?  If we call it Validation but treat it effectively as a second set of actions that run after the Configure step.  This would let you write Validations that both Assert/SetDefaults.  What you have today I think would be good as something built on top, sugar that makes it easy for you to register an expression that returns a bool. 

Something like `Validate(Action<TOptions> options)` as well.

What do you think?

MaKCbIMKo: So, you suggest adding a few overloads for `Validate` method with a lightweight signature, correct?

I was thinking about using `Action<TOptions>` as a parameter. But it would require to use exceptions to signal that options are invalid. As far as I remember, using bool is faster than exceptions. Moreover, if will expect exception we will have to wrap into try/catch block each separated validation call (in other case validation will be stopped after first validation failed).

I was trying to use `Func<TOptions, bool>` at first, but I found `Expression` more informative, because it provides information that might help identify the problem (and might be logged).

HaoK: Yeah I was thinking we layer things on.  So at the lowest layer there's a Validate that just takes an Action, then we can build additional overloads that take expression/func and return bools, that way there's still the flexibility to drop down and do something different if they don't want to follow the more constrained func/expression/bool route.  

We wouldn't necessarily have to do anything to support wrapping exceptions either this way, since that's the lowest level API

HaoK:
Take a look at https://github.com/aspnet/Options/issues/161 which is tracking some features which would be useful for options to provide for the Auth 2.0 work, named options will add a bit of a wrinkle since there'd need to be the ability to target All instances vs named instances, vs the default instance.  We haven't decided that we are bringing back named options yet, but its something to consider

src/Microsoft.Extensions.Options.Validation/OptionsValidationServiceCollectionExtensions.cs
MaKCbIMKo: I've got your point. I will add the lowest layer overloads.

HaoK: I'm thinking maybe the higher level overload should be `Func<TOptions, IEnumerable<Exception>>`  they can return multiple validation exceptions this way so its slightly more powerful than a bool

MaKCbIMKo:
@HaoK, I've modified (a lot) of validation stuff. Moreover, I took a look on feature (about multi-tenant) and found that it might require some validation 'on creation' time. I also added that (wasn't not a big deal).

About multi-tenancy: it all depends on how we will configure all these 'names', where we will store cached instances (for different names), but I think it won't be so difficult to modify current validation to work in a new way.

But, only if you have some roadblocks that you still not sure (or don't know) how to break...

HaoK:
Thanks, I'll try to take a look this week

HaoK:
So I'm going to be starting a PR for https://github.com/aspnet/Options/issues/161, it will be a pretty lightweight feature to start, I'll submit it by EOD tomorrow, so you can provide input and insight since you've been playing around with validation for a while.  The initial critical feature is named options support, so the first iteration of the PR will likely have very minimal validation support (think `Action<TOption>`).

MaKCbIMKo:
Ok, waiting for your PR.

HaoK:
Take a look at https://github.com/aspnet/Options/pull/176

MaKCbIMKo:
@HaoK - Is it fine to base this PR over your initial Options 2.0 changes? Or there more changes are comming?

Maybe something in my PR need to be reconsidered...

HaoK:
Feel free to rebase if you want, there likely will be more changes coming, but it won't be for a few weeks

HaoK:
See https://github.com/aspnet/Options/pull/187, adding an Initialize set of actions that runs after Configure for fixup, you could potentially use these for validation as well.  Auth is going to run validation outside of options since it needs to run every request potentially, so it will only use this feature for fixup

MaKCbIMKo:
@HaoK - At first view, 'Initialization' step might be used for validation, but this validation will be applied all the time. Is that fine? What if want to validate only one time at start up (is that real case ?) ?

MaKCbIMKo:
update:  I implemented a common validation (for unnamed options).

Faced some troubles with named options: If we want to validate the options at start up, then we need to know all 'names'. It means that we need to store somewhere all names. But it argues with ability to inject the options that wasn't configured at all (because of `IOptions` default feature). 
Still thinking about that.

MaKCbIMKo:
@HaoK - I have added a possibility to validate all named options. Please take a look.

p.s. there is some problems with Travis CI check (Error: Unable to locate libunwind. Install libunwind to continue) - not sure how to solve that weather the problem with my changes.

MaKCbIMKo:
Changes from dev branch has been merged. @HaoK - may I ask you to review the PR?

MaKCbIMKo:
About 'AppVeyor build failed' - seems like it faced the exception during downloading "https://dotnetcli.azureedge.net/dotnet/Sdk/2.0.0-preview3-006785/dotnet-sdk-2.0.0-preview3-006785-win-x64.zip" (I managed to download it locally). then it switched to alt download link "https://dotnetcli.azureedge.net/dotnet/Sdk/2.0.0-preview3-006785/dotnet-dev-win-x64.2.0.0-preview3-006785.zip" - which seems to be not working.

@HaoK - maybe we just need to try to re-run the AppVeyor build?

MaKCbIMKo:
@HaoK - did you have a chance to take a look at this PR?

HaoK:
I took a look, but we haven't signed up to do this work yet, so drilling into the details/implementation is a bit premature until we actually are trying to get this feature implemented.

MaKCbIMKo:
Do you mean that this feature is not in roadmap?

Can we do anything in order to sign this up?

HaoK:
Yes, feedback/comments in https://github.com/aspnet/Options/issues/227 saying why you want/need this feature will help get it prioritized higher/sooner.

aspnet-hello:
This issue was moved to aspnet/Home#2392

aspnet-hello:
Sorry I was a bad bot. Re-opening this PR.

HaoK:
See https://github.com/aspnet/Options/pull/266 for initial validation support

Issue #185, opened 2017-04-20
Add back Initialize


HaoK:
Add a `Initialize<TOptions>` set of actions which run after Configure, needed for Auth

MaKCbIMKo:
A quick question: Why `Initialize` goes after `Configure`? Does it correlate with principle of least astonishment? By these names, seems like you need to initialize object first and then configure that.

HaoK:
Yeah its a naming issue, I'm open to considering different names if you have any ideas for something clearer that signify's happening after Configure.

MaKCbIMKo:
I suppose, that just switch the names isn't option? :)

How about that:
First step - `Configure` is like a 'rough processing'.
Second step - `Adjust` or `[Fine]tune` for polishing (fine tuning, polishing)

HaoK:
52ab5f16e7c358a27f4affd14da2ef7082d3d6d1

PR #187, opened 2017-04-26
Add Initialize<TOptions>

HaoK:
https://github.com/aspnet/Options/issues/185

Needed to replace double locking/options fixup in auth as we discussed last week.

@ajcvickers @Tratcher @Eilon @davidfowl

### DIFF
--- a/src/Microsoft.Extensions.Options/InitializeOptions.cs
+++ b/src/Microsoft.Extensions.Options/InitializeOptions.cs
@@ -1,0 +1,54 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Extensions.Options
+{
+    /// <summary>
+    /// Implementation of IInitializeNamedOptions.
+    /// </summary>
+    /// <typeparam name="TOptions"></typeparam>
+    public class InitializeOptions<TOptions> : IInitializeOptions<TOptions> where TOptions : class
+    {
+        /// <summary>
+        /// Constructor.
+        /// </summary>
+        /// <param name="name">The name of the options.</param>
+        /// <param name="action">The action to register.</param>
+        public InitializeOptions(string name, Action<TOptions> action)
+        {
+            Name = name;
+            Action = action;
+        }
+
+        /// <summary>
+        /// The options name.
+        /// </summary>
+        public string Name { get; }
+
+        /// <summary>
+        /// The configuration action.
+        /// </summary>
+        public Action<TOptions> Action { get; }
+
+        /// <summary>
+        /// Invokes the registered Initialize Action if the name matches.
+        /// </summary>
+        /// <param name="name"></param>
+        /// <param name="options"></param>
+        public virtual void Initialize(string name, TOptions options)
+        {
+            if (options == null)
+            {
+                throw new ArgumentNullException(nameof(options));
+            }
+
+            // Null name is used to Initialize all named options.
+            if (Name == null || name == Name)
+            {
+                Action?.Invoke(options);
+            }
+        }
+    }
+}
--- a/src/Microsoft.Extensions.Options/InitializeOptions.cs
+++ b/src/Microsoft.Extensions.Options/InitializeOptions.cs
@@ -6,13 +6,13 @@ using System;
 namespace Microsoft.Extensions.Options
 {
     /// <summary>
-    /// Implementation of IInitializeNamedOptions.
+    /// Implementation of <see cref="IInitializeOptions{TOptions}"/>.
     /// </summary>
     /// <typeparam name="TOptions"></typeparam>
     public class InitializeOptions<TOptions> : IInitializeOptions<TOptions> where TOptions : class
     {
         /// <summary>
-        /// Constructor.
+        /// Creates a new instance of <see cref="InitializeOptions{TOptions}"/>.
         /// </summary>
         /// <param name="name">The name of the options.</param>
         /// <param name="action">The action to register.</param>
--- a/src/Microsoft.Extensions.Options/InitializeOptions.cs
+++ b/src/Microsoft.Extensions.Options/InitializeOptions.cs
@@ -28,12 +28,12 @@ namespace Microsoft.Extensions.Options
         public string Name { get; }
 
         /// <summary>
-        /// The configuration action.
+        /// The initialization action.
         /// </summary>
         public Action<TOptions> Action { get; }
 
         /// <summary>
-        /// Invokes the registered Initialize Action if the name matches.
+        /// Invokes the registered initialization Action if the name matches.
         /// </summary>
         /// <param name="name"></param>
         /// <param name="options"></param>
@@ -44,7 +44,7 @@ namespace Microsoft.Extensions.Options
                 throw new ArgumentNullException(nameof(options));
             }
 
-            // Null name is used to Initialize all named options.
+            // Null name is used to initialize all named options.
             if (Name == null || name == Name)
             {
                 Action?.Invoke(options);
--- a/src/Microsoft.Extensions.Options/OptionsFactory.cs
+++ b/src/Microsoft.Extensions.Options/OptionsFactory.cs
@@ -12,14 +12,17 @@ namespace Microsoft.Extensions.Options
     public class OptionsFactory<TOptions> : IOptionsFactory<TOptions> where TOptions : class, new()
     {
         private readonly IEnumerable<IConfigureOptions<TOptions>> _setups;
+        private readonly IEnumerable<IInitializeOptions<TOptions>> _initializers;
 
         /// <summary>
         /// Initializes a new instance with the specified options configurations.
         /// </summary>
         /// <param name="setups">The configuration actions to run.</param>
-        public OptionsFactory(IEnumerable<IConfigureOptions<TOptions>> setups)
+        /// <param name="initializers">The initialization actions to run.</param>
+        public OptionsFactory(IEnumerable<IConfigureOptions<TOptions>> setups, IEnumerable<IInitializeOptions<TOptions>> initializers)
         {
             _setups = setups;
+            _initializers = initializers;
         }
 
         public TOptions Create(string name)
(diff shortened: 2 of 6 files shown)


MaKCbIMKo:
For me it looks like the same as `Configure` step (does the same). But another step right after the configuration one. Is the main goal of that is make sure that it will be called right after all configurations?

What if something will require step number 3 (`PostConfigure` or something else)?

Maybe need to create more complex solution? I know that's too much, but more long-term solution.

src/Microsoft.Extensions.Options/OptionsFactory.cs
```diff
@@ -12,14 +12,17 @@ namespace Microsoft.Extensions.Options
     public class OptionsFactory<TOptions> : IOptionsFactory<TOptions> where TOptions : class, new()
     {
         private readonly IEnumerable<IConfigureOptions<TOptions>> _setups;
+        private readonly IEnumerable<IInitializeOptions<TOptions>> _initializers;
 
         /// <summary>
         /// Initializes a new instance with the specified options configurations.
         /// </summary>
         /// <param name="setups">The configuration actions to run.</param>
-        public OptionsFactory(IEnumerable<IConfigureOptions<TOptions>> setups)
```
Eilon: Should we keep the old ctor for compat, and have it chain to the new ctor? DI would only call the new ctor anyway, but it would avoid a breaking change.

src/Microsoft.Extensions.Options/OptionsServiceCollectionExtensions.cs
```diff
@@ -69,5 +69,43 @@ public static IServiceCollection Configure<TOptions>(this IServiceCollection ser
 
         public static IServiceCollection ConfigureAll<TOptions>(this IServiceCollection services, Action<TOptions> configureOptions) where TOptions : class
             => services.Configure(name: null, configureOptions: configureOptions);
+
+        /// <summary>
+        /// Registers an action used to Initialize a particular type of options.
```
Eilon: Don't need to capitalize the `I` of `Initialize` (it's just a word here, not a method name).

Eilon: (Applies to other doc comments in this PR.)

src/Microsoft.Extensions.Options/OptionsServiceCollectionExtensions.cs
```diff
@@ -69,5 +69,43 @@ public static IServiceCollection Configure<TOptions>(this IServiceCollection ser
...
+        /// </summary>
+        /// <typeparam name="TOptions">The options type to be Initialized.</typeparam>
+        /// <param name="services">The <see cref="IServiceCollection"/> to add the services to.</param>
+        /// <param name="name">The name of the options instance.</param>
+        /// <param name="InitializeOptions">The action used to Initialize the options.</param>
+        /// <returns>The <see cref="IServiceCollection"/> so that additional calls can be chained.</returns>
+        public static IServiceCollection Initialize<TOptions>(this IServiceCollection services, string name, Action<TOptions> InitializeOptions)
```
Eilon: Rename `InitializeOptions` param to start with lowercase letter.

src/Microsoft.Extensions.Options/InitializeOptions.cs
```diff
@@ -0,0 +1,54 @@
...
+namespace Microsoft.Extensions.Options
+{
+    /// <summary>
+    /// Implementation of IInitializeNamedOptions.
+    /// </summary>
+    /// <typeparam name="TOptions"></typeparam>
+    public class InitializeOptions<TOptions> : IInitializeOptions<TOptions> where TOptions : class
+    {
+        /// <summary>
+        /// Constructor.
```
Eilon: `Creates a new instance of...`

src/Microsoft.Extensions.Options/InitializeOptions.cs
```diff
@@ -0,0 +1,54 @@
...
+        /// </summary>
+        /// <param name="name">The name of the options.</param>
+        /// <param name="action">The action to register.</param>
+        public InitializeOptions(string name, Action<TOptions> action)
+        {
+            Name = name;
```
Eilon: Do these params need any checks? Can they be empty? Null? Anything?

src/Microsoft.Extensions.Options/InitializeOptions.cs
```diff
@@ -0,0 +1,54 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Extensions.Options
+{
+    /// <summary>
+    /// Implementation of IInitializeNamedOptions.
```
Eilon: Can use a typeref for the type in this doc.

test/Microsoft.Extensions.Options.Test/OptionsSnapshotTest.cs
```diff
@@ -209,6 +209,85 @@ public void CanConfigureAllDefaultAndNamedOptions()
...
 
         [Fact]
+        public void CanConfigureAndInitializeAllDefaultAndNamedOptions()
+        {
+            var services = new ServiceCollection().AddOptions();
+            services.ConfigureAll<FakeOptions>(o => o.Message += "Default");
+            services.Configure<FakeOptions>(o => o.Message += "0");
+            services.Configure<FakeOptions>("1", o => o.Message += "1");
+            services.InitializeAll<FakeOptions>(o => o.Message += "Initialize");
+            services.Initialize<FakeOptions>(o => o.Message += "2");
```
Eilon: I think I'm just confused, but doesn't this `Initialize` initialize *all* FakeOptions, because no name was specified??

src/Microsoft.Extensions.Options/OptionsFactory.cs
HaoK: We added this type in 2.0  so no breaking change thankfully

src/Microsoft.Extensions.Options/OptionsServiceCollectionExtensions.cs
HaoK: Yeah I'll fix, global find/replace fail, I just cloned the ConfigureOptions files and replaced

src/Microsoft.Extensions.Options/InitializeOptions.cs
HaoK: null name signify's applying to all names.
string.empty is used to target the 'default' name, i.e. the overloads that don't specify name

Passing in a null action is equivalent to passing in a no-op o => { }, so I don't think its something we explicitly need to block (we didn't for the existing https://github.com/aspnet/Options/blob/dev/src/Microsoft.Extensions.Options/ConfigureOptions.cs, so we need to be consistent)  Note: we do require non null actions on the sugar `Configure/Initialize` apis since it doesn't make sense to add no-op/null actions there.

test/Microsoft.Extensions.Options.Test/OptionsSnapshotTest.cs
HaoK: Yeah its a bit confusing,

All overloads that don't take a name target the Options.DefaultName = string.Empty;  This was done to make all of the existing options.Configures behave the same in Options 2.0 since we added names, and wanted to preserve the existing semantics and behaviors.

ajcvickers: I think we should define some non-empty string to have this meaning. Like "Default", or something more specific if we think avoiding collisions is important. The empty-string is very un-documenting.

src/Microsoft.Extensions.Options/InitializeOptions.cs
Eilon: Ok.

test/Microsoft.Extensions.Options.Test/OptionsSnapshotTest.cs
Eilon: Yeah let's get that figured out.

HaoK: Well, the actual value for the default name doesn't matter too much, we use Options.DefaultName everywhere, its currently set to string.empty since that seemed like a good default to me, in general the constant should be used everywhere but if we want to change it, its just changing this line: https://github.com/aspnet/Options/blob/dev/src/Microsoft.Extensions.Options/Options.cs#L11

HaoK: To clarify, whatever value we pick, its just used as the default for the overloads that don't specify a name, so it doesn't really have any documention value other than if you were debugging a custom IConfigure/InitializeOptions.

HaoK: I think its pretty clear today if you look at the usage in the overloads with the constant:

https://github.com/aspnet/Options/blob/dev/src/Microsoft.Extensions.Options/OptionsServiceCollectionExtensions.cs#L43

Eilon: I don't have a strong opinion on the exact name, I just feel strongly about "let's get that figured out" 😄 @ajcvickers ?

ajcvickers: I don't feel strongly, so I'll let @HaoK make the call. I would be a little confused if I was attempting to debug code that was using this, or if I passed empty string and go back something, but I guess most people won't do those things.

HaoK: Okay lets keep it as empty string for now, since this isn't new for Initialize.

empty string is pretty weird name to pick, and you wouldn't actually see any issues with using it as your name, unless you did configured using both methods.

services.Configure(o => o.Foo = "bar");
services.Configure("", o => o.Foo = "what??");

HaoK:
Updated to fix casing, also added references in the doc comments for Configure/Initialize to each other and whether they run before/after each other.

Also added a test/fix for a minor issue with old IConfigureOptions being run for all named instances instead of only the default instance

HaoK:
Merged 52ab5f16e7c358a27f4affd14da2ef7082d3d6d1

PR #189, opened 2017-05-22
Add infrastructure needed for config defaults

HaoK:
Needed for https://github.com/aspnet/MetaPackages/issues/117

@ajcvickers

### DIFF
--- a/src/Microsoft.Extensions.Options/Infrastructure/ConfigureDefaultOptions.cs
+++ b/src/Microsoft.Extensions.Options/Infrastructure/ConfigureDefaultOptions.cs
@@ -1,0 +1,66 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Extensions.Options.Infrastructure
+{
+    /// <summary>
+    /// Used by ConfigureDefaults to configure defaults.
+    /// </summary>
+    /// <typeparam name="TOptions"></typeparam>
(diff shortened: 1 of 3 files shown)


src/Microsoft.Extensions.Options/Infrastructure/ConfigureDefaultOptions.cs
```diff
@@ -0,0 +1,66 @@
...
+
+        /// <summary>
+        /// The configuration action.
+        /// </summary>
+        public Action<TOptions> Action { get; }
+
+        /// <summary>
+        /// Invokes the registered configure Action if the name matches.
+        /// </summary>
+        /// <param name="name"></param>
```
ajcvickers: Missing docs

src/Microsoft.Extensions.Options/Infrastructure/ConfigureDefaultOptions.cs
```diff
@@ -0,0 +1,66 @@
...
+
+using System;
+
+namespace Microsoft.Extensions.Options.Infrastructure
+{
+    /// <summary>
+    /// Used by ConfigureDefaults to configure defaults.
+    /// </summary>
+    /// <typeparam name="TOptions"></typeparam>
+    public class ConfigureDefaultOptions<TOptions> where TOptions : class
```
pakrym: Why if this class and not interface?

HaoK: We wanted to make it ghetto and something people don't use

Issue #191, opened 2017-06-01
Rename IInitializeOptions to IPostConfigureOptions


danroth27:
@divega @HaoK @ajcvickers 

Both `IConfigureOptions` and `IInitializeOptions` have similar semantics but they are called at different times. Based on the names it sounds like `IInitializeOptions` would happen before `IConfigureOptions`, but apparently the order is the opposite. Can we come up with a better name that makes the behavior more obvious?

HaoK:
Yeah don't disagree, need a name that conveys this is a `PostConfigure` phase.

danroth27:
What were the scenarios that motivated adding this additional stage? The scenarios might help determine the name. I believe you said one of the scenarios was validation?

HaoK:
Its basically Fixup/Validation, but we didn't want to bite off the entire validation feature for options, so we targeted only fixup/normalization, auth uses it to do things to establish defaults, in your azure options for example, I use this to establish set the authority

HaoK:
So if we want an uber lame name that's extremely clear, `service.PostConfigure<TOptions>()` does fit the description.

danroth27:
@glennc

divega:
Is this about triggering all the `IConfigureOptions` to run?

HaoK:
867583d945f0297212ddc11d6863562d31cb251d

divega:
A few of us talked about this with @davidfowl and he believes we really should revisit this decision.

HaoK:
For post preview2 right?

HaoK:
@divega can you summarize what we want to revisit, just the naming or its existence? :)

divega:
@HaoK very briefly, @davidfowl thinks that the name `IPostConfigureOptions` is bad and wants us to think harder about it.

divega:
@HaoK could you please summarize what `IPostConfigureOptions` is for?

HaoK:
It should be thought of as a parallel set of Configure's that behave identically, but all run after Configure, in the order they were registered.  

The common scenario will be to calculate defaults from what's been already configuration.

HaoK:
All that said PostConfigure is explicitly correct, any intent we give it is fairly arbitrary, its basically Configure2

HaoK:
But something like [I]Calculate/Compute[Options] might imply running after Configure better than Initialize...

divega:
[...]

PR #192, opened 2017-06-02
Intialize => PostConfigure

HaoK:
Fixes https://github.com/aspnet/Options/issues/191

Will need to update security as well.

cc @ajcvickers @divega


HaoK:
867583d945f0297212ddc11d6863562d31cb251d

Issue #199, opened 2017-06-14
IOptionsChangeTokenSource does not work with scoped IConfigureOptions


pakrym:
Not sure if this is intended behavior but it is confusing. 
Consider the following example:

``` csharp
 public class Startup
    {
        // This method gets called by the runtime. Use this method to add services to the container.
        // For more information on how to configure your application, visit https://go.microsoft.com/fwlink/?LinkID=398940
        public void ConfigureServices(IServiceCollection services)
        {
            services.AddScoped<IConfigureOptions<ScopedOptions>, ScopedConfigureOptions>();
        }

        // This method gets called by the runtime. Use this method to configure the HTTP request pipeline.
        public void Configure(IApplicationBuilder app, IHostingEnvironment env)
        {
            if (env.IsDevelopment())
            {
                app.UseDeveloperExceptionPage();
            }

            app.Run(async (context) =>
            {
                var optionsSnapshot = context.RequestServices.GetService<IOptionsSnapshot<ScopedOptions>>();
                await context.Response.WriteAsync(optionsSnapshot.Value.I.ToString());
            });
        }
    }

    public class ScopedConfigureOptions : IConfigureOptions<ScopedOptions>
    {
        private int _i;
        private static int _global;

        public ScopedConfigureOptions()
        {
            _i = _global++;
        }

        public void Configure(ScopedOptions options)
        {
            options.I = _i;
        }
    }

    public class ScopedOptions
    {
        public int I { get; set; }
    }
```

I would expect `OptionsSnapshot` being scoped to initialize options instance using new instance of scoped `IConfigureOptions` every time instead last first one is cached forever (if `IOptionsChangeTokenSource` is not used)

@HaoK

pakrym:
Scoped IOptionsChangeTokenSource on the other hand are supported

HaoK:
Yeah this is a bug, the cache is singleton, so even though the snapshot is recreated, the cache'd value is returned.

The snapshot shouldn't be using the IOptionsCache at all, we should consider switching IOptionsMonitor over to the IOptionsCache/IOptionsFactory and having that be the official new way to get live options instances + do cache invaldations

HaoK:
Basically OptionsSnapshot should only be using the factory, and the instance itself is the scoped cache.

HaoK:
There's a second issue here as a result, snapshot shouldn't really be doing the changeSource registration.  I'll file an issue for that now

PinpointTownes:
[...]

PR #202, opened 2017-06-15
Fix scoped snapshot bug

HaoK:
Fixes https://github.com/aspnet/Options/issues/199

cc @ajcvickers @pakrym @divega


pakrym:
Can you please add a test verifies scenario from the bug works? Do we need IOptionsCache as abstraction now?

HaoK:
I started with your test, and merged with the existing one after making sure it passed.

RE IOptionsCache, depends on if we have time to update IOptionsMonitor for 2.0, if we don't update it, we will probably have to end up adding basically an IOptionsMonitor2 in 2.1 to unify things.

Issue #201, opened 2017-06-15
IOptionsMonitor should be updated for 2.0


HaoK:
Related https://github.com/aspnet/Options/issues/199

Snapshot shouldn't responsible for registering IOptionsChangeTokenSource

I think we need to update IOptionsMonitor to use the new 2.0 functionality especially now that logging uses it.

Proposed changes for `IOptionsMonitor` 
- Use IOptionsFactory / support names / post configures
- Use the IOptionsCache to get live updates / handle cache invalidations (should potentially revisit the config reload token hookup to see if there's any improvements there)

@ajcvickers @divega thoughts?  With logging using options monitor, its gone from not very important to super important scenario, and it got 0 love in 2.0 so far...

HaoK:
Note: if we don't end up updating IOptionsMonitor to use IOptionsCache in 2.0, we should remove the public  IOptionsCache since it won't be used anywhere now that Snapshot isn't using it. cc @pakrym

divega:
@HaoK I like the idea because if I understand correctly this also helps making things more consistent across all options flavors. It adds to the work that you need to do for 2.0, and I am not sure if that is a good idea :smile:

HaoK:
@divega and I chatted and we're going to try and make options consistent for 2.0.

IOptions/IOptionsMonitor should both switch to using IOptionsFactory to pickup IPostConfigureOptions.

We should support named options at a minimum for IOptionsMonitor and maybe IOptions if I have time

HaoK:
50a900bbb01690463db0a4aa4ce755bb7c7100e1

Issue #197, opened 2017-06-14
Should IOptions also consume IPostConfigureOptions


HaoK:
See https://github.com/aspnet/Options/issues/191

@divega @davidfowl @ajcvickers

HaoK:
50a900bbb01690463db0a4aa4ce755bb7c7100e1

PR #203, opened 2017-06-19
Normalize Options[Monitor] to use IOptionsFactory

HaoK:
All 3 variantations of IOptions now consistently use IOptionsFactory

Standardizes Names/PostConfigure behavior and support. (With the exception of IOptions not supporting names, we can add INamedOptions later to avoid breaking existing IOptions implementations)

Single shared OptionsManager used to implement IOptions + IOptionsSnapshot.

Fixes https://github.com/aspnet/Options/issues/201 https://github.com/aspnet/Options/issues/197

@divega @ajcvickers @davidfowl @pakrym


HaoK:
Notifcations/IOptionsChangeTokenSource still needs to be updated to support names

src/Microsoft.Extensions.Options/OptionsMonitor.cs
```diff
@@ -13,21 +13,22 @@ namespace Microsoft.Extensions.Options
     /// <typeparam name="TOptions"></typeparam>
     public class OptionsMonitor<TOptions> : IOptionsMonitor<TOptions> where TOptions : class, new()
     {
-        private LegacyOptionsCache<TOptions> _optionsCache;
-        private readonly IEnumerable<IConfigureOptions<TOptions>> _setups;
+        private readonly IOptionsCache<TOptions> _cache;
```
HaoK: IOptionsMonitorCache

HaoK:
50a900bbb01690463db0a4aa4ce755bb7c7100e1

Issue #204, opened 2017-06-20
IOptionsMonitor is singleton but listener registration is not thread safe


pakrym:
Bad things might happen if two consumers subscribe concurrently. 
Also if subscription happens during invocation.
/cc @HaoK

HaoK:
Yuck, yup not threadsafe

HaoK:
Discussed with @ajcvickers will fix with either using a `SynchronizedCollection` or explicit locks

pakrym:
You can also use `event` internally it is already synchronized.

HaoK:
Good idea, this is pretty much the exact usage pattern for events...

HaoK:
ff4200632970c53a3a6c860219d10477d74060d6

PR #209, opened 2017-06-26
Switch to internal event for Monitor

HaoK:
@ajcvickers @pakrym 

Fixes https://github.com/aspnet/Options/issues/204

### DIFF
--- a/src/Microsoft.Extensions.Options/OptionsMonitor.cs
+++ b/src/Microsoft.Extensions.Options/OptionsMonitor.cs
@@ -16,7 +16,7 @@ namespace Microsoft.Extensions.Options
         private readonly IOptionsMonitorCache<TOptions> _cache;
         private readonly IOptionsFactory<TOptions> _factory;
         private readonly IEnumerable<IOptionsChangeTokenSource<TOptions>> _sources;
-        internal event EventHandler<ChangeArgs> _onChange;
+        internal event Action<TOptions> _onChange;
 
         /// <summary>
         /// Constructor.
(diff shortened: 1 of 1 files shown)


src/Microsoft.Extensions.Options/OptionsMonitor.cs
```diff
@@ -16,7 +16,7 @@ public class OptionsMonitor<TOptions> : IOptionsMonitor<TOptions> where TOptions
         private readonly IOptionsMonitorCache<TOptions> _cache;
         private readonly IOptionsFactory<TOptions> _factory;
         private readonly IEnumerable<IOptionsChangeTokenSource<TOptions>> _sources;
-        private List<Action<TOptions>> _listeners = new List<Action<TOptions>>();
+        internal event EventHandler<ChangeArgs> _onChange;
```
pakrym: Event can be `Action` too

HaoK:
Updated to use `Action<TOptions>` which cleans things up a bit

src/Microsoft.Extensions.Options/OptionsMonitor.cs
```diff
@@ -72,26 +72,32 @@ public virtual TOptions Get(string name)
...
         public IDisposable OnChange(Action<TOptions> listener)
         {
-            var disposable = new ChangeTrackerDisposable(_listeners, listener);
-            _listeners.Add(listener);
+            var disposable = new ChangeTrackerDisposable(this, listener);
+            _onChange += disposable.OnChange;
             return disposable;
         }
 
+        internal class ChangeArgs : EventArgs
```
pakrym: Don't need anymore

HaoK:
ff4200632970c53a3a6c860219d10477d74060d6

PR #212, opened 2017-07-04
Add name support for Monitor callbacks

HaoK:
@ajcvickers @pakrym @divega 

Action<TOptions, string> is kinda ugly, maybe we should make it an OptionsCallback class or something?

### DIFF
--- a/src/Microsoft.Extensions.Options/IOptionsMonitor.cs
+++ b/src/Microsoft.Extensions.Options/IOptionsMonitor.cs
@@ -26,6 +26,6 @@ namespace Microsoft.Extensions.Options
         /// </summary>
         /// <param name="listener">The action to be invoked when TOptions has changed.</param>
         /// <returns>An IDisposable which should be disposed to stop listening for changes.</returns>
-        IDisposable OnChange(Action<TOptions> listener);
+        IDisposable OnChange(Action<TOptions,string> listener);
     }
 }
--- a/src/Microsoft.Extensions.Options/IOptionsMonitor.cs
+++ b/src/Microsoft.Extensions.Options/IOptionsMonitor.cs
@@ -22,7 +22,7 @@ namespace Microsoft.Extensions.Options
         TOptions Get(string name);
 
         /// <summary>
-        /// Registers a listener to be called whenever TOptions changes.
+        /// Registers a listener to be called whenever a named TOptions changes.
         /// </summary>
         /// <param name="listener">The action to be invoked when TOptions has changed.</param>
         /// <returns>An IDisposable which should be disposed to stop listening for changes.</returns>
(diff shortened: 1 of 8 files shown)


src/Microsoft.Extensions.Options/IOptionsMonitor.cs
```diff
@@ -22,10 +22,10 @@ public interface IOptionsMonitor<out TOptions>
         TOptions Get(string name);
 
         /// <summary>
-        /// Registers a listener to be called whenever TOptions changes.
+        /// Registers a listener to be called whenever a named TOptions changes.
         /// </summary>
         /// <param name="listener">The action to be invoked when TOptions has changed.</param>
         /// <returns>An IDisposable which should be disposed to stop listening for changes.</returns>
-        IDisposable OnChange(Action<TOptions> listener);
+        IDisposable OnChange(Action<TOptions,string> listener);
```
pakrym: nit; space

pakrym:
Imagine a situation when we have `IConfiguration` and based on requested options name we read section from it. How would I need to setup change tokens to reset all options read from this IConfiguration without knowing names in front?

HaoK:
We could add a clear all to IOptionsMonitorCache, it was something I considered before...

HaoK:
Updated adding Clear() to IOptionsMonitorCache which drops all options instances

pakrym:
So how do I write `IOptionsChangeTokenSource` that clears all options from cache now?

HaoK:
You should be able to just take IOptionsMonitorCache in your Source and call Clear no?  The cache has no dependencies.

pakrym:
It's strange that we have two asymmetric, why have change token at all then? Each source can inject cache and call clean with the name.

HaoK:
The changetokens are there to fire the notifications,, invalidating the cache does not cause any updates/notifications to listeners, while change tokens will trigger the listeners immediately.

pakrym:
Back to my original question then, how do I trigger token for all named instances without registering token source for each?

My problem is that `ConfigureOptions` and `IOptionsChangeTokenSource` come in pairs but while one ConfigureOptions can support arbitrary number of named option objects `IOptionsChangeTokenSource` needs to be registered per each. My proposal is to add name argument to `GetChangeToken` method.

HaoK:
So you want to register a single IOptionsChangeTokenSource that generates the appropriate change token for a named options instance via GetChangeToken(name)? But existing token sources can continue to just ignore this flag

HaoK:
So looking at how things are right now, we expect people to register a change token source per named option they watch, so IChangeTokenSource.Name shows that.  

At this point I don't think we have the runway to add a new path that enables a change token source that is able to generate events for more.

I think we'll have to wait for 2.1 to add support for that, I'll discuss with @ajcvickers today this PR to make the call on what we have in here.

HaoK:
After discussing with @ajcvickers we'll go with what we have here, I'll file an issue for us to consider the scenario you described with a single change token source responsible for multiple names.

PR #219, opened 2017-07-19
Add ServiceProvider aware overloads to Configure

couven92:
Provides overloads to the `Configure` methods that do not require an actual `IConfiguration` instance, but accept a delegate that enables `IConfiguration` to be retrieved from the `IServiceProvider` when `IOptions<TOptions>` is requested.

However, this requires `IConfiguration` to be registered in the service container.


couven92:
Alternatively, the configAccessor argument could be changed to a `Func<IServiceProvider, IConfiguration>` delegate instead. This might give people a better idea of what this overload is for, as it would line up with the `AddSingleton` (and related) extension methods.

HaoK:
This should no longer be needed, you should be able to do something similar via the new Configure overload that hangs off OptionsBuilder that accepts generic dependencies, 

see edc21af4166574ecd45d8f8dbd381dfec044f367

Issue #222, opened 2017-08-14
Extension methods for more concise registration of IConfigureOptions/IPostConfigureOptions


tuespetre:
`services.AddSingleton<IConfigureOptions<MvcOptions>, MyConfigureMvcOptions>()` seems unwieldy and non-obvious. I believe a friendlier, more discoverable alternative would be `services.Configure<MvcOptions, MyConfigureMvcOptions>()`.

```
public static IServiceCollection Configure<TOptions, TConfigureOptions>(this IServiceCollection services)
    where TOptions : class
    where TConfigureOptions : class, IConfigureOptions<TOptions>
{
    return services.AddSingleton<IConfigureOptions<TOptions>, TConfigureOptions>();
}

public static IServiceCollection Configure<TOptions, TPostConfigureOptions>(this IServiceCollection services)
    where TOptions : class
    where TPostConfigureOptions: class, IPostConfigureOptions<TOptions>
{
    return services.AddSingleton<IPostConfigureOptions<TOptions>, TPostConfigureOptions>();
}
```

HaoK:
We used to have sugar for this early in 1.0 which we removed near 1.0 RTM, we actually allowed registration of a global setup class, and would reflect on it registering all IConfigureOptions it implemented.

i.e.

```
  services.RegisterOptionsSetup<MyConfigure>(); // Reflects and registers all of the appropriate

  class MyConfigure : IConfigureOptions<MvcOptions>, IConfigureOptions<RazorOptions>, IPostConfigureOptions<MvcOptions> {
     // implement everything you want in one class
  }
```

We could bring something similar back

Issue #224, opened 2017-08-17
Semi-Automatic options configuration improvements


HaoK:
From @dougbu @davidfowl 

A couple of months back, Fowler and I discussed ways to make it easier to configure options. My main concern at that time was the machinations necessary to set options based on other services – the whole `IConfigureOptions<TOptions>` rigmarole feels at best second class. We’re hoping you can find a germ in this worth adding sooner or later. That is, please run with something.

I wrote up the https://gist.github.com/dougbu/bc4880f0aadce9d94b7df91d53782da0 gist based on these discussions. The basic thing is semi-automatic discovery of `Startup` methods that configure options. These methods are all named `ConfigureOptions`, take a first parameter of the `TOptions` they set, and may have additional parameters grabbed from DI.

David and I covered a few questions but didn’t completely resolve them: 
-	Should the `ConfigureOptions` methods always be found and executed? Or, must the user call `services.Configure<TOptions>()` for every `TOptions` or perhaps a less-granular `services.ConfigureOptions()`?
-	Must `ConfigureOptions` methods be within `Startup`? For example, the explicit configuration request might be `services.ConfigureOptions(this)` if you do want to use the `Startup` class.
-	When are the `ConfigureOptions` methods executed and in what relative order? Granular `services.Configure<TOptions>()` methods might be semantically identical to placing an `IConfigureOptions<TOptions>` implementation into DI at that time. If supported, we’d need to decide if `services.ConfigureOptions()` called the methods in whatever order Reflection returns, alphabetically by `TOptions` name, or something else. And if `UseStartup<Startup>` implies `ConfigureOptions()` discovery, are the methods executed before or after `IConfigureOptions<TOptions>` implementations that `ConfigureServices()` adds?

About all we resolved is that what we have now isn’t great and `Startup: IConfigureOptions<MvcOptions>, IConfigureOptions<RazorViewEngineOptions>` wouldn’t help with service resolution (besides being butt-ugly).

HaoK:
This is somewhat related to https://github.com/aspnet/Options/issues/222

Assuming we had something like a `services.RegisterOptionsSetup<T>` which would reflect on T and register all of the various IConfigureOptions that T implements.  We could then just use that new api to register Startup automatically, the one difference from your gist would be that Startup would need to implement the interfaces.  Or maybe we should just have a new dedicated SetupOptions class for this, and register it by convention if it exists:

```C#
[...]

PR #230, opened 2017-09-05
Add services.ConfigureOptions<T>(Type) overloads

HaoK:
Fixes https://github.com/aspnet/Options/issues/222

Adds a few new ways to configure options:
- ConfigureOptions(Type)
- `ConfigureOptions<TSetup>()`
- ConfigureOptions(object setupInstance)

All of these use reflection to register all of the
IConfigureOptions/IPostConfigureOptions on the type/instance

Note: I ressurrected these from what we had back in 1.0.0-beta8 (not sure about the names)

Also will be used for: https://github.com/aspnet/Options/issues/224

cc @davidfowl @ajcvickers @divega


src/Microsoft.Extensions.Options/ConfigureNamedOptions.cs
```diff
@@ -9,7 +9,7 @@ namespace Microsoft.Extensions.Options
     /// Implementation of IConfigureNamedOptions.
     /// </summary>
     /// <typeparam name="TOptions"></typeparam>
-    public class ConfigureNamedOptions<TOptions> : IConfigureNamedOptions<TOptions>, IConfigureOptions<TOptions> where TOptions : class
+    public class ConfigureNamedOptions<TOptions> : IConfigureNamedOptions<TOptions> where TOptions : class
```
davidfowl: Isn't this a breaking change?

HaoK: IConfigureNamedOptions derives from IConfigureOptions, this is just cleanup

davidfowl:
@HaoK should this throw if you implement both interfaces? What's the expected behavior then?

HaoK:
So basically to do named options in a non breaking way, how it works is that IConfigureOptions are treated as IConfigureNamedOptions with name == Options.DefaultName.  

So even though they have to implement both interfaces for IConfigureNamedOptions, but the old `Configure(TOptions o)` is never called

HaoK:
The correct implementation is always to forward like this: `Configure(TOptions o) => Configure(Options.DefaultName, o)`

Issue #220, opened 2017-07-19
Inconsistent use of null vs DefaultName


Tratcher:
With introduction of named options there are a series of APIs that require you to pass in the name of your instance. However, there is inconsistent behavior for working with the default instance. E.g. `IOptionsFactory<T>.Create(name)` accepts `null` to create the default instance, but `IOptionsMonitor<T>.TryRemove(name)` throws for null and requires you to use `Options.DefaultName` for the default instance.  Decide if nulls are allowed or not, but be consistent. If they are, then silently map to DefaultName as appropriate.

HaoK:
We should probably relax things and just use the default name instead of throwing anywhere when a null name is specified.  

Note we do use null name in a few cases to mean ignore the name, and target all named instances (i.e. ConfigureNamedOptions)

PR #231, opened 2017-09-05
Allow null name to mean default instance

HaoK:
Fixes https://github.com/aspnet/Options/issues/220

@ajcvickers @Tratcher


HaoK:
The only danger with this change, is here:

`services.Configure(name: null, o => o.DoWhatever())` means configure all instances regardless of name
`IOptionsMonitor.Get(name: null)` means give me the Options.DefaultInstance

I don't think this is too big of a deal, but its worth noting that where's just trading a different set of inconsistencies.

HaoK:
I'm not opposed to keeping things as they are, where null is allowed in only a few places, and in general it throws.  I don't think this defaulting behavior will be generally useful.

Tratcher:
It's the inconsistency that bothers me. This PR at least addresses that.

HaoK:
I guess that's fair, we won't throw for null names anywhere

Issue #237, opened 2017-09-30
OptionsBuilder to chain configuration calls


jdom:
Hi, I'm not sure if there is already built-in support for this scenario, so maybe the suggestion is irrelevant, but we'd like to have services added to the service collection, as well as configure them in the process.
Today this would look like:

```
services.AddTransient<IFooService, FooServices>();
services.Configure<FooServiceOptions>(myConfiguration);
services.Configure<FooServiceOptions>(options => ...);
services.Configure<FooServiceOptions>(options => ...);

// then add logic so that an IHostedService initializes an instead of FooService
```

For our users, we could create an extension method to add the `FooService` as well as specify different overloads to be able to configure in all the different ways that the Options infrastructure support, but that means A TON of extension method overloads, and even adding a reference to `Microsoft.Extensions.Options.ConfigurationExtensions` just in case the user happens to use that. Also if some external library adds more overloads to `.Configure`, we can't support those.

Alternatively, one can create an extension method that just adds FooService to the collection, and then the user has to explicitly call `Configure<FooServiceOptions>` multiple times. The main problem here is discoverability: how does the user know that he needed to configure `FooServiceOptions` when using `FooService`.
This approach also explodes when we talk about named options.

This challenge could be overcome if the Options package provides an `IOptionsBuilder<TOptions>` interface that allows calls to `Configure` to be chained. Only if the user references `Microsoft.Extensions.Options.ConfigurationExtensions` then the `.Configure(IConfiguration) overload would light up, and so on.

This way our users should be able to write:

```
services.AddFooService("myName")
    .Configure(myConfiguration)
    .Configure(options => ...);
```

Implementation should be very straightforward: `OptionsBuilder<TOptions>` would just have to hold the name that we are configuring, as well as a reference to IServiceCollection, so that it can forward the calls to it.

Does it make sense? I'm thinking of adding infrastructure like this to configure the many services in Orleans, but ideally this would be part of the main Options package. I can send a PR if you think it's a good idea.

xiazen:
This is a more flexible pattern which would be very good to have. And it would benefit all applications/software which tries to integrate with `Options `

jdom:
Any comments about this? Is this a valid approach or is there another way to achieve this that I missed?
 Even if we don't want to do this in this repo for now, do you think it would be OK if I implement it in the Orleans infrastructure?

HaoK:
[...]

PR #238, opened 2017-10-09
Options builder to chain Configure calls of the same named TOptions

jdom:
Addresses https://github.com/aspnet/Options/issues/237

This allows to add a named service to the application, as well as allow the end-user to configure its options, without the need to discover what are the TOptions it needs to chain calls to, nor having to pass the name on every .`Configure<TOptions>(name, ...)` call.

This results in the user being able to do things like this for extensions that use the builder:

```
services.AddNamedXyz("name", optionsBuilder => 
   optionsBuilder.Configure(Configuration.GetSection("Something"))
   optionsBuilder.Configure(options => options.SomeSetting = "value for name"));
```

Notice how the 2 calls to the Configure method automatically infer the `TOptions` type, and do not need to pass the name (even thought they are configuring a named option.

Also, the author of the `AddNamedXyz` extension method does not need to reference the `Microsoft.Extensions.Options.ConfigurationExtension` nuget package just in case the end user cares to use the `IConfiguration` overload. If the end user references it, then the relevant `Configure` overload will automatically light up on the options builder.


src/Microsoft.Extensions.Options/IOptionsBuilder.cs
```diff
@@ -0,0 +1,24 @@
...
+
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Microsoft.Extensions.Options
+{
+    /// <summary>
+    /// Represents a builder that configures a named TOptions instance.
+    /// </summary>
+    /// <typeparam name="TOptions">The type of options being requested.</typeparam>
+    public interface IOptionsBuilder<TOptions> where TOptions : class
```
jdom: Is `IOptionsBuilder` a good name? Because it is meant to forward calls to the `IServiceCollection`. An alternative is to make it agnostic of it, and then have the concrete implementation `flush` the calls to the service collection, but I would think that's a little less intuitive in case you mix using the builder and using Configure directly on the service collection

test/Microsoft.Extensions.Options.Test/OptionsBuilderTests.cs
```diff
@@ -0,0 +1,58 @@
...
+
+namespace Microsoft.Extensions.Options.Tests
+{
+    public class OptionsBuilderTest
+    {
+        [Fact]
+        public void CanSupportDefaultName()
+        {
+            var services = new ServiceCollection().AddOptions();
+            var builder = services.CreateOptionsBuilder<FakeOptions>();
```
HaoK: Nit: maybe shorten `CreateOptionsBuilder` => `BuildOptions`?

src/Microsoft.Extensions.Options/IOptionsBuilder.cs
HaoK: I'd prefer if the only thing you can do with this builder is call Configure as opposed to allowing arbitrary access to IServiceCollection.

i.e. the only operations that would be allowed on the builder would be Configure/PostConfigure (or anything else specifically targeting the builder)

HaoK: Is there any major benefit to having an interface for this as opposed to only the OptionsBuilder class itself, you can still add extension methods in the same way.  There should be nothing that is worth replacing in the base class really.

HaoK: As this is all sugar on top of existing `IServiceCollection` methods

jdom: There isn't really a benefit for the interface. As it is, it's just syntactic sugar on top of `IServiceCollection`, which is the main challenge I'm addressing. I started this thinking I would store delegates and then have an explicit call that would bind them to the `IServiceCollection`, but that's what I though would be un-intuitive with no real gain (other than decoupling the builder from the DI container temporarily, which IMO is not useful).
So I can change this to be just a public type for us to hook up extension methods.
Still, the specific extension methods would need to access the `IServiceCollection` to be able to forward calls to it, or do you suggest some other approach?

HaoK: Yeah that's fine, we typically hang onto the IServiceCollection in our builders too, and you will definitely have to hang onto the Name as well.  

The main benefit I see is strongly tying a specific named options instance to this builder, so you can just call Configure/PostConfigure without the generic or specifying the name.

I'm still trying to see if we could use this instead of things like `ConfigureApplicationCookie` that we have today.

jdom: Hehe, configuring the cookie is the only example I saw of using named options in the wild, and I'm not terribly fond of the approach, and hence this proposal. In Orleans we'll have many cases of configuring named services/options.
But even for our singleton service abstractions that can be fulfilled with 1 of different implementations, where each of them might have slightly different typed options, this can be of a lot of use, as the end user will not have to discover which strongly typed options to use, as well as giving the flexibility to chain multiple calls (and not just pick only 1 of the `Action<TOptions>` or `IConfiguration` overloads, in case we provide the extension methods with these overloads).

src/Microsoft.Extensions.Options/OptionsServiceCollectionExtensions.cs
```diff
@@ -189,5 +189,32 @@ public static IServiceCollection ConfigureOptions(this IServiceCollection servic
...
             return services;
         }
+
+        /// <summary>
+        /// Creates an options builder that forwards Configure calls for the same <typeparamref name="TOptions"/> to the underlying service collection.
+        /// </summary>
+        /// <typeparam name="TOptions">The options type to be configured.</typeparam>
+        /// <param name="services">The <see cref="IServiceCollection"/> to add the services to.</param>
+        /// <returns>The <see cref="OptionsBuilder{TOptions}"/> so that configure calls can be chained in it.</returns>
+        public static OptionsBuilder<TOptions> BuildOptions<TOptions>(this IServiceCollection services) where TOptions : class
```
davidfowl: I'm not sure about this name.

jdom: What about these alternatives?
1. `Configure<TOptions>()` (as it would be the only of these overloads that don't take in an extra argument, but it might be confusing that they have different return values)
2. `CreateOptionsBuilder<TOptions>()`
3. `ConfigureOptions<TOptions>()` (although there is already a non-generic overload with this name that takes a type, so it might be very confusing)
4. `BuildOptions<TOptions>()` (just putting it here for contrasting)
5. `ConfiguratorFor<TOptions>()` (probably might make sense to rename `OptionsBuilder` to `OptionsConfigurator` if we choose this extension method name, and maybe even regardless, as this helper doesn't build from scratch, just helps configure).
6. `UseConfigurator<TOptions>()`

jdom:
Any preferences for naming, or further feedback on the approach?

src/Microsoft.Extensions.Options/OptionsServiceCollectionExtensions.cs
jdom: BTW, reflecting on this, my preference would be option 5.

HaoK:
So I thought about this a bit more, and I think we can maybe do something more drastic, akin to what we did with authentication...

Where we add `OptionsBuilder`, which has `OptionsBuilder<T> Build<T>(string name, Action<OptionsBuilder<T>> build)` method.

So startup would look more like:
```
  services.BuildOptions()
    .Build<Options1>("name", o => o.Configure(config)) // named instance
    .Build<Options1>(o => o.Configure(config)) // no name
```

This lets us scope options apis to OptionsBuilder and not pollute IServiceCollection, and we can still hang the apis specific to a particular options type (hiding the name for example) to an instance of `OptionsBuilder<T>`

Thoughts @davidfowl @ajcvickers @jdom ?

jdom:
@HaoK your suggestion will work for us, so we'll gladly take it, although I'm not sure it is better IMO. The current approach just adds a single extension method to IServiceCollection, with an overload for taking the name (so technically 2 methods with the same name).
Your suggestion has only 1 total, but you have to chain it to be useful, otherwise BuildOptions() by itself doesn't add value. Also, at least in our scenarios, it's not that the end user will start configuring a bunch of named options in one go... they will do it while they are adding concrete service instances, so it makes sense that they they get the options builder/configurator at the time they add that specific service. For that I mean they'll do this:
```
services
    .AddFooService("service1", ob => ob.Configure(config).Configure(o => o.FooSetting = 3))
    .AddBarService("service2", ob => ob.Configure(config).Configure(o => o.BarSetting = "bar"))
    .AddBarService("service3", ob => ob.Configure(config));
```
While your approach would favor more this:
```
services
    .AddFooService("service1")
    .AddBarService("service2")
    .AddBarService("service3")
    .BuildOptions()
         .Build<FooServiceOptions>("service1", ob => ob.Configure(config).Configure(o => o.FooSetting = 3)
         .Build<BarServiceOptions>("service2", ob => ob.Configure(config).Configure(o => o.BarSetting = "bar"))
         .Build<BarServiceOptions>("service3", ob => ob.Configure(config)));
```
Granted, it doesn't prevent the original code, but you can see that where it's most useful is not by chaining different option builders together, but chaining the addition of services with their options, and the user doesn't need to discover which options to configure, or what name to correlate with the named service.

HaoK:
So I'd expect you to just have `AddFooService` return `OptionsBuilder<FooOptions>`, and then you'd still be able to do something like:

`services.AddFooService("service1").Configure(config).Configure(o => o.FooSetting = 3))`

HaoK:
The `BuildOptions` block would be more for if you wanted to configure all the options in once place.  `AddXyz` should still be free (and encouraged) to return the appropriate sugar to make it easy to configure things.

HaoK:
@jdom I'm going to prototype the OptionsBuilder in this PR: https://github.com/aspnet/Options/pull/239 as we can have these new Configure overloads return an `OptionsBuilder<T>` instead of IServiceCollection

HaoK:
Looks good to me, any concerns @davidfowl @ajcvickers ?

src/Microsoft.Extensions.Options.ConfigurationExtensions/OptionsBuilderConfigurationExtensions.cs
```diff
@@ -0,0 +1,33 @@
...
+    public static class OptionsBuilderConfigurationExtensions
+    {
+        /// <summary>
+        /// Registers a configuration instance which TOptions will bind against.
+        /// </summary>
+        /// <typeparam name="TOptions">The options type to be configured.</typeparam>
+        /// <param name="optionsBuilder">The options builder to add the services to.</param>
+        /// <param name="config">The configuration being bound.</param>
+        /// <returns>The <see cref="OptionsBuilder{TOptions}"/> so that additional calls can be chained.</returns>
+        public static OptionsBuilder<TOptions> Configure<TOptions>(this OptionsBuilder<TOptions> optionsBuilder, IConfiguration config) where TOptions : class
```
HaoK: Actually now that we have a Builder to hang this off of, we don't have to call it Configure, we should call this method `Bind` to give it more meaning

jdom: I would gladly do that change, but let's clarify the bigger picture first. Does this also apply to the overload that takes and `Action<TOptions>`. And would we rename `PostConfigure` to `PostBind`?
Or maybe those overloads should be renamed to `Apply` and `PostApply`?  Or `Update`?

HaoK: Yeah we probably should switch to something different, its going to be too insane to keep adding all these overloads as longer term I wouldn't be surprised if there were more lifecycles either, like Validation.

Which probably would mean something closer to:

Bind(IConfiguration, OptionsLifecycle.[Configure/PostConfigure/Validation]).

To level we'd still have
Configure
PostConfigure
(Validate in the future)

HaoK: It looks like we don't have PostConfigure(IConfiguration) today anyways, so I don't think you need to worry about it for the purposes of this PR...  We can tackle this in a different issue/PR

HaoK: Its just sugar for `AddOptions<T>.PostConfigure(o => config.Bind(o))` anyways.

ajcvickers:
@HaoK Looks okay to me.

HaoK:
Thanks @jdom I'll merge this into dev now

Issue #236, opened 2017-09-29
Add Configure<TOptions, TService1...5>(Action<TOptions>, TService1...5) overloads


HaoK:
See discussion here: https://github.com/aspnet/StaticFiles/issues/184#issuecomment-333186413

Basic jist is to add sugar for a transient configure options that takes dependencies.

We would have to do this for named + postConfigure variants as well.

`services.Configure<TOptions, TDependency>(Action<TOptions> configure)`

which would register as transient something like:
```
class ConfigureOptions<TOptions, TDependency> : IConfigureOptions<TOptions> {
    ConfigureWithDI(TDependency dep) => Dep = dep;
    TDependency Dep { get; }
}
```

HaoK:
edc21af4166574ecd45d8f8dbd381dfec044f367

PR #239, opened 2017-10-16
Add Configure<TOption, TDep1, TDep2> sugar

HaoK:
For https://github.com/aspnet/Options/issues/236

Making sure the current pattern looks good before adding all of the rest boilderplate (will also add matching PostConfigure versions...

Basic idea will be to enable: (up to 5??)

These register a transient func that just uses GetRequiredService for each Dep, its up to the caller to add the deps themselves with whatever lifetime they want.

```
services.Configure<TOptions, Dep1, Dep2, Dep3, Dep4, Dep5>("namedOption", o,d1,d2,d3,d4,d5 => o.Something = d1.A + d2.B + d3.C + d4.D + d5.E)
```

@davidfowl @ajcvickers


khellang:
Is there a reason you won't just add a metod taking `Action<TOptions, IServiceProvider>` that gets wrapped in an `IConfigureOptions<TOptions>` and registered in the container?

davidfowl:
> Is there a reason you won't just add a metod taking Action<TOptions, IServiceProvider> that gets wrapped in an IConfigureOptions<TOptions> and registered in the container?

We should just do both as part of this PR. May as well bring this back https://github.com/aspnet/Options/pull/152.

src/Microsoft.Extensions.Options/ConfigureNamedOptions.cs
```diff
@@ -53,4 +53,177 @@ public virtual void Configure(string name, TOptions options)
...
+        }
+
+        public void Configure(TOptions options) => Configure(Options.DefaultName, options);
+    }
+
+    /// <summary>
+    /// Implementation of IConfigureOptions.
+    /// </summary>
+    /// <typeparam name="TOptions"></typeparam>
+    /// <typeparam name="TDep"></typeparam>
```
HaoK: Dep1, Dep2...5

HaoK:
Updated PR adding Configure and PostConfigure methods taking 0 to 5 dependencies.  Also added `OptionsBuilder<TOptions>`  to potentially be the place to hang these overloads... so the builder binds to a specific named instance for chaining purposes.

Examples:
```C#
            var default = services.BuildOptions<FakeOptions>()
                   .Configure<SomeService>((o,s) => o.Message = s.Stuff);

            var named = services.BuildOptions<FakeOptions>("named")
                   .Configure<SomeService>((o, s) => o.Message = "named "+s.Stuff);

            // Future framework methods can also expose their options builders
            var xyzOptionsBuilder = services.AddXyz(o => o.Set = "Something")
                    .Configure(o, s => o.Other= s.Compute())
```

In terms of naming, we could go with `services.AddOptions<TOptions>` instead of Build since we currently don't use that overload

jdom:
I do think that the OptionsBuilder is the right place to put all of these extensions. I would keep the original `Configure<TOptions>` on IServiceCollection, but only for back compatibility, but in theory they could just forward to this. The only ones that don't really fit in this builder are the `ConfigureAll` extension methods, as these apply to all, and not just to a named option.

BTW, as I mentioned in https://github.com/aspnet/Options/pull/238#discussion_r144434431, I'm not that fond of calling it `Builder` or use `Build` in the name (as I originally did), as it is not really building something isolated from IServiceCollection, but instead it's helping configure the collection**. I think `OptionsConfigurator` is a better name for the class, but haven't decided on the extension method. Maybe `AddOptions<T>` as you suggested is good. The ideal would be `ConfigureOptions<T>`, but that's already taken.

** what I mean it's not building in isolation, is that these 2 calls would effectively configure the same options one after the other, and not build 2 different instances of the options, of course:
```
services.BuildOptions<FakeOptions>().Configure(o => o.Message = "1");
// ... 
services.BuildOptions<FakeOptions>().Configure(o => o.Message += "2");
```

HaoK:
Yeah AddOptions might work better since it behaves like the other AddXyz()'s today where if you call it twice the options manipulations are just run in order:

```
            var default = services.AddOptions<FakeOptions>()
                   .Configure<SomeService>((o,s) => o.Message = s.Stuff);

            var named = services.AddOptions<FakeOptions>("named")
                   .Configure<SomeService>((o, s) => o.Message = "named "+s.Stuff);
```

jdom:
Cool, I'll update my OptionsBuilder PR with the new names:
```
OptionsBuilder => OptionsConfigurator
BuildOptions<TOptions> extension method => AddOptions<TOptions> extension method
```
I will also include the `PostConfigure` method that I left out for my initial PR.

HaoK:
We can leave the class name alone, we already have this pattern every where (MvcBuilder/IdentityBuilder/AuthenticationBuilder), all of them are basically adding sugar methods for manipulating the IServiceCollection

jdom:
Hah, OK, I'll revert that commit then, I was about to push :)

HaoK:
This should be ready to review now for real, builds on top of the new OptionsBuilder we added via @jdom 's PR:
- [Post]Configure<D1...D5> were added to IOptionsBuilder
- Test added to ensure IServiceProvider works as a dependency as well

HaoK:
edc21af4166574ecd45d8f8dbd381dfec044f367

PR #251, opened 2018-03-14
Create or update baselines to match 2.0

ryanbrandenburg:


src/Microsoft.Extensions.Options/breakingchanges.netcore.json
```diff
@@ -1,22 +1,12 @@
 [
     {
-      "TypeId": "public class Microsoft.Extensions.Options.OptionsManager<T0> : Microsoft.Extensions.Options.IOptions<T0> where T0 : class, new()",
-      "MemberId": "public .ctor(System.Collections.Generic.IEnumerable<Microsoft.Extensions.Options.IConfigureOptions<T0>> setups)",
-      "Kind": "Removal"
+        "TypeId": "public class Microsoft.Extensions.Options.OptionsWrapper<T0> : Microsoft.Extensions.Options.IOptions<T0> where T0 : class, new()",
```
javiercn: @HaoK Is this a breaking change?

HaoK: Technically yes, but given that the 2.0 implementation threw NotImplemented exception and the method doesn't exist on an interface...  https://github.com/aspnet/Options/blob/rel/2.0.0/src/Microsoft.Extensions.Options/OptionsWrapper.cs#L37

Its fine I think

javiercn: Looping @Eilon as he's the decider of all things breaking changes

Eilon: Yeah I don't think this is an acceptable breaking change. We didn't even mark it as obsolete or doc it as such. I recommend we add back the missing API and continue to throw. We can certainly follow the guidance at https://github.com/aspnet/Home/wiki/Engineering-guidelines#breaking-changes and delete it in the future.

I logged https://github.com/aspnet/Home/issues/2983 to fix the breaking change.

@ryanbrandenburg - I'm OK with checking this in as you have it, then when we fix https://github.com/aspnet/Home/issues/2983 we can revert this breaking change entry.

Eilon: cc @ajcvickers here as well.

PR #255, opened 2018-03-26
Undo breaking change to OptionsWrapper

HaoK:
Fixes: https://github.com/aspnet/Home/issues/2983

Filed https://github.com/aspnet/Home/issues/3002 to remove these in 3.0


ajcvickers (approved):

Eilon:
Can we mark them as obsolete per the guidelines mentioned in https://github.com/aspnet/Home/issues/2983 ?

HaoK:
Is that warning really necessary in this specific case given that its non virtual and throws if its ever called?

Eilon:
I think it'd be good to follow the guidelines, because at least the docs will be correct for anyone looking at the API refs, though I agree it's surprising if anyone in practice was calling it.

HaoK:
Okay if the docs will reflect it, that's pretty useful, I"ll send another PR

PR #256, opened 2018-03-26
Obsolete wrapper methods

HaoK:
Also obsoletes `Get` which is also an orphaned method from IOptionsMonitor which OptionsWrapper does not implement.


Eilon (approved):
Nice and easy. Thanks!

PR #258, opened 2018-05-27
Make OptionsFactory.Create virtual

hikalkan:
I want to derive from this class and override it.

### DIFF
--- a/src/Microsoft.Extensions.Options/OptionsFactory.cs
+++ b/src/Microsoft.Extensions.Options/OptionsFactory.cs
@@ -25,7 +25,7 @@ namespace Microsoft.Extensions.Options
             _postConfigures = postConfigures;
         }
 
-        public TOptions Create(string name)
+        public virtual TOptions Create(string name)
         {
             var options = new TOptions();
             foreach (var setup in _setups)


hikalkan:
Hi,
Can you accept this please?
Thanks.

Eilon:
Looks like this is considered a breaking change: https://github.com/MichalStrehovsky/corefx/blob/master/Documentation/coding-guidelines/breaking-change-rules.md#members

> ## Members
> ...
> X Disallowed
> ...
> Adding virtual to a member
>
> While this change would often work without breaking too many scenarios because C# compiler tends to emit callvirt IL instructions to call non-virtual methods (callvirt performs a null check, while a normal call won't), we can't rely on it. C# is not the only language we target and the C# compiler increasingly tries to optimize callvirt to a normal call whenever the target method is non-virtual and the this is provably not null (such as a method accessed through the ?. null propagation operator). Making a method virtual would mean that consumer code would often end up calling it non-virtually.

Eilon:
@hikalkan - can you share more info about your scenario? Maybe we can come up with an alternative approach?

hikalkan:
Making this would be a breaking change in the IL level, I understand. But you can add this to the next version where people will re-compile their projects on upgrade.

Anyway, this is my requirement:

I wanted to create a dedicated OptionsFactory implementation to create `IdentityOptions` of the AspNet Core Identity library. Why needed?

I'm building a multi-tenant application (actually, an application framework: https://github.com/aspnetboilerplate/aspnetboilerplate) where **every tenant will have its own Identity options** (like Password complexity, user lockout or signin settings). As you know, when you inject `IOptions<IdentityOptions>`, all instances are same and it's singleton (shared for the application lifetime).

So, my solution is that:

1. Register it scoped, instead of singleton for the `IdentityOptions`:

````C#
services.Replace(ServiceDescriptor.Scoped<IOptions<IdentityOptions>, OptionsManager<IdentityOptions>>());
````

2. Create a special OptionsFactory implementation (`AbpIdentityOptionsFactory` here) that dynamically determine the option values based on the current tenant:

````C#
services.Replace(ServiceDescriptor.Transient<IOptionsFactory<IdentityOptions>, AbpIdentityOptionsFactory>());
````

The implemantation:

````C#
public class AbpIdentityOptionsFactory : AbpOptionsFactory<IdentityOptions>
{
    private readonly ISettingManager _settingManager;

    public AbpIdentityOptionsFactory(
        IEnumerable<IConfigureOptions<IdentityOptions>> setups,
        IEnumerable<IPostConfigureOptions<IdentityOptions>> postConfigures,
        ISettingManager settingManager)
        : base(setups, postConfigures)
    {
        _settingManager = settingManager;
    }

    public override IdentityOptions Create(string name)
    {
        var options = base.Create(name);

        OverrideOptions(options);

        return options;
    }

    protected virtual void OverrideOptions(IdentityOptions options)
    {
        options.Password.RequiredLength = _settingManager.Get(IdentitySettingNames.Password.RequiredLength, options.Password.RequiredLength);
        //...override other options like that... (_settingManager gets settings based on the current tenant)
    }
}
````

This implementation perfectly solves the problem. Any class injects `IOptions<IdentityOptions>` gets the custom options for the current tenant.

`AbpIdentityOptionsFactory` uses `base.Create(name);`. So, it needs that the base.Create method should be `virtual`.

#### Additional comments

BTW, I suggest you to use virtuals more in the framework classes to make them easily overridable. Otherwise, we need to copy/paste some code from your framework to ours :)

You may find this dangerous if people overrides it and breaks an aspnet core functionality. But breaking a functionality is easier than we generally think. So, don't worry about that much and make the framework as easily extensible and customizable as possible. 99% of the people will not even use it. But it will allow us, framework/library developers, more options.

natemcmaster:
I'm closing this PR as we have decided to merge this repo into https://github.com/aspnet/Extensions.

If you feel this is something that is still important, feel free to re-open a PR against the new repo.

PR #266, opened 2018-06-29
Add support for options validation

HaoK:
Adds the ability to validate options via a new `IValidateOptions` interface that the `IOptionsFactory` consumes if registered:

General use case looks like this and is only exposed off of the `OptionsBuilder`:

```C#
            services.AddOptions<ComplexOptions>()
                .Configure(o =>
                {
                    o.Boolean = false;
                    o.Integer = 11;
                })
                .Validate(o => o.Boolean)
                .Validate(o => o.Integer > 12);
            var error = Assert.Throws<OptionsValidationException>(() => sp.GetRequiredService<IOptions<ComplexOptions>>().Value);
            var errors = error.Failures.ToArray();
            Assert.Equal(2, errors.Length);
            Assert.Equal("A validation error has occured.", errors[0]);
            Assert.Equal("A validation error has occured.", errors[1]);

```

Thoughts @davidfowl ?

I've already gone over the general idea with @ajcvickers and we think it looks promising.

### DIFF
--- a/src/Microsoft.Extensions.Options/IValidateOptions.cs
+++ b/src/Microsoft.Extensions.Options/IValidateOptions.cs
@@ -1,0 +1,20 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.Extensions.Options
+{
+    /// <summary>
+    /// Interface used to validate options.
+    /// </summary>
+    /// <typeparam name="TOptions">The options type to validate.</typeparam>
+    public interface IValidateOptions<TOptions> where TOptions : class
+    {
+        /// <summary>
+        /// Validates a specific named options instance (or all when name is null).
+        /// </summary>
+        /// <param name="name">The name of the options instance being validated.</param>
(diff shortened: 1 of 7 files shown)


davidfowl:
The feature looks generally useful. What's the need for 2.2? Which components are going to use it?

ajcvickers:
@davidfowl Let's discuss in the leads sync today.

HaoK:
There isn't an explicit need for this in 2.2 as opposed to 3.0, Security/Auth would probably want to switch to using this in some form to validate options.  

The other thing that we should try to tackle as part of this is enabling some way to trigger validation at startup instead of on first use, maybe via IStartupFilter like this blog post: https://andrewlock.net/adding-validation-to-strongly-typed-configuration-objects-in-asp-net-core/

HaoK:
Addresses https://github.com/aspnet/Home/issues/2388 which I've moved into 2.2

HaoK:
Gonna put this into 2.2, and look at triggering start up validation in a separate PR

davidfowl:
cc @rynowak @anurse

rynowak:
@davidfowl @HaoK It's neat, is there a suggestion about any MVC features where we'd use it?

PR #272, opened 2018-08-07
Add DataAnnotations based validation

HaoK:
Add DataAnnotations validator support: (https://github.com/aspnet/Home/issues/3385)

Basic support works pretty much as expected:

```C#
        [AttributeUsage(AttributeTargets.Field | AttributeTargets.Property, AllowMultiple = false, Inherited = true)]
        public class FromAttribute : ValidationAttribute
        {
            public string Accepted { get; set; }

            public override bool IsValid(object value)
                => value == null || value.ToString() == Accepted;
        }

        private class AnnotatedOptions
        {
            [Required]
            public string Required { get; set; }

            [StringLength(5, ErrorMessage = "Too long.")]
            public string StringLength { get; set; }

            [Range(-5, 5, ErrorMessage = "Out of range.")]
            public int IntRange { get; set; }

            [From(Accepted = "USA")]
            public string Custom { get; set; }
        }

        [Fact]
        public void CanValidateDataAnnotations()
        {
            services.AddOptions<AnnotatedOptions>()
                .Configure(o =>
                {
                    o.StringLength = "111111";
                    o.IntRange = 10;
                    o.Custom = "nowhere";
                })
                .ValidateDataAnnotations();

            var error = Assert.Throws<OptionsValidationException>(() => sp.GetRequiredService<IOptions<AnnotatedOptions>>().Value);
            ValidateFailure<AnnotatedOptions>(error, Options.DefaultName, "The Required field is required. : Too long. : Out of range. : The field Custom is invalid.");
        }
```

Basically you get a single error message with all of the data validation errors smushed together, since its considered a single options validator.

Thoughts? @ajcvickers @davidfowl @Eilon @DamianEdwards


HaoK:
We can split this out into a separate Options.DataAnnotations package if we still care about keeping dependencies of the core options package lean...

ajcvickers:
Code seems good, but not sure about taking a dependency on DataAnnotations--we probably shouldn't do it now unless we want to keep it in 3.0. But if other things already have the dependency and will continue to have it in 3.0, then their probably won't be a problem.

davidfowl:
No not in core options. This should be another package reference unlesss we make it part of net standard

DamianEdwards:
+1 for new package layer

HaoK:
`Microsoft.Extensions.Options.DataAnnotations` for the new package name?

src/Microsoft.Extensions.Options.DataAnnotations/DataAnnotationValidateOptions.cs
```diff
@@ -0,0 +1,50 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.ComponentModel.DataAnnotations;
+using System.Linq;
+
+namespace Microsoft.Extensions.Options
+{
+    /// <summary>
+    /// Implementation of <see cref="IValidateOptions{TOptions}"/>
```
Eilon: Can maybe elaborate on this a bit more? This comment is just stating what can be seen from the type declaration anyway.

src/Microsoft.Extensions.Options.DataAnnotations/DataAnnotationValidateOptions.cs
```diff
@@ -0,0 +1,50 @@
...
+    /// </summary>
+    /// <typeparam name="TOptions">The instance being validated.</typeparam>
+    public class DataAnnotationValidateOptions<TOptions> : IValidateOptions<TOptions> where TOptions : class
+    {
+        public DataAnnotationValidateOptions(string name)
```
Eilon: Need doc comments on all public APIs.

src/Microsoft.Extensions.Options.DataAnnotations/OptionsBuilderDataAnnotationsExtensions.cs
```diff
@@ -0,0 +1,26 @@
...
+
+namespace Microsoft.Extensions.DependencyInjection
+{
+    /// <summary>
+    /// Extension methods for adding configuration related options services to the DI container via <see cref="OptionsBuilder{TOptions}"/>.
+    /// </summary>
+    public static class OptionsBuilderDataAnnotationsExtensions
+    {
+        /// <summary>
+        /// Register this options instance for the built in DataAnnotationsAttribute
```
Eilon: The term `DataAnnotationsAttribute` looks like it's being used as a type name, but it isn't a type. Rephrase somehow?

HaoK:
Fixed new doc comments and removed the doc comment excludes so added all the missing doc comments as well

src/Microsoft.Extensions.Options.DataAnnotations/DataAnnotationValidateOptions.cs
```diff
@@ -0,0 +1,60 @@
...
+                var vc = new ValidationContext(options, serviceProvider: null, items: null);
+                var isValid = Validator.TryValidateObject(options, vc, validationResults, validateAllProperties: true);
+
+                if (isValid)
+                {
+                    return ValidateOptionsResult.Success;
+                }
+
+
+                var errors = String.Join(" : ", validationResults.Select(r => r.ErrorMessage));
```
Eilon: This is a bit icky. I'm guessing there's no "list of errors" that we can send a list of these failures to? How about put one on each line instead?

```
"Validation of the object XYZ failed:
{reason 1}
{reason 2}
{reason 3}"
```

Eilon: I think it would be also useful here to use ValidationResult.MemberNames and format that in some way so that the user knows which property(s) failed.

src/Microsoft.Extensions.Options.ConfigurationExtensions/Microsoft.Extensions.Options.ConfigurationExtensions.csproj
```diff
@@ -3,7 +3,7 @@
   <PropertyGroup>
     <Description>Provides additional configuration specific functionality related to Options.</Description>
     <TargetFramework>netstandard2.0</TargetFramework>
-    <NoWarn>$(NoWarn);CS1591</NoWarn>
```
Eilon: YES! Another project finally has full docs!

src/Microsoft.Extensions.Options.DataAnnotations/DataAnnotationValidateOptions.cs
HaoK: Okay I'll try to clean this up, yeah the data annotation validations are treated as a single validaiton failure inside of an arbitrary amount of other validators, which is why it gets turned into a single string, since it potentially would need to be aggregated with other non data-annotation based validation errors.

Eilon: Yeah I think at least putting 1 per line would be a good separation.

HaoK: Specifically only for data annotations errors? or all of the aggregated errors as well?  I'll add a new test that demonstrates what it looks like so its easier to review in the PR

Eilon: Yeah let's see it in the test then I'll review.

HaoK:
Updated the new errors are more wordy now but hopefully more clear:
```
"DataAnnotation validation failed for members {Required} with error 'The Required field is required.'.
DataAnnotation validation failed for members {StringLength} with error 'Too long.'.
DataAnnotation validation failed for members {IntRange} with error 'Out of range.'.
DataAnnotation validation failed for members {Custom} with error 'The field Custom is invalid.'."
```

src/Microsoft.Extensions.Options.DataAnnotations/DataAnnotationValidateOptions.cs
```diff
@@ -49,7 +49,10 @@ public ValidateOptionsResult Validate(string name, TOptions options)
                 }
 
 
-                var errors = String.Join(" : ", validationResults.Select(r => r.ErrorMessage));
+                var errors = String.Join(Environment.NewLine, 
+                    validationResults.Select(r => "DataAnnotation validation failed for members {" +
```
Eilon: Maybe get rid of the `{`? A bit too nerdy 😄

src/Microsoft.Extensions.Options.DataAnnotations/DataAnnotationValidateOptions.cs
```diff
@@ -49,7 +49,10 @@ public ValidateOptionsResult Validate(string name, TOptions options)
                 }
 
 
-                var errors = String.Join(" : ", validationResults.Select(r => r.ErrorMessage));
+                var errors = String.Join(Environment.NewLine, 
+                    validationResults.Select(r => "DataAnnotation validation failed for members {" + 
+                    String.Join(",", r.MemberNames) + 
+                    "} with error '" + r.ErrorMessage + "'."));
```
Eilon: ` with *the* error `

test/Microsoft.Extensions.Options.Test/OptionsBuilderTest.cs
```diff
@@ -555,7 +555,39 @@ public void CanValidateDataAnnotations()
...
+            var sp = services.BuildServiceProvider();
+
+            var error = Assert.Throws<OptionsValidationException>(() => sp.GetRequiredService<IOptions<AnnotatedOptions>>().Value);
+            ValidateFailure<AnnotatedOptions>(error, Options.DefaultName, 
+                @"DataAnnotation validation failed for members {Required} with error 'The Required field is required.'.
+DataAnnotation validation failed for members {StringLength} with error 'Too long.'.
+DataAnnotation validation failed for members {IntRange} with error 'Out of range.'.
+DataAnnotation validation failed for members {Custom} with error 'The field Custom is invalid.'.",
+                "I don't want to go to nowhere!");
+        }
```
Eilon: Could we add a case where there is more than 1 member as well? Maybe CompareValidator does that? Or write a dummy custom validator that does that?

HaoK:
Just as FYI, multi member error looks like:

DataAnnotation validation failed for members Dep1,Dep2 with the error 'Dep1 != Dep2'.

src/Microsoft.Extensions.Options.DataAnnotations/DataAnnotationValidateOptions.cs
```diff
@@ -40,20 +40,18 @@ public ValidateOptionsResult Validate(string name, TOptions options)
...
 
-
-                var errors = String.Join(Environment.NewLine, 
-                    validationResults.Select(r => "DataAnnotation validation failed for members {" + 
-                    String.Join(",", r.MemberNames) + 
-                    "} with error '" + r.ErrorMessage + "'."));
-                return ValidateOptionsResult.Fail(errors);
+                return ValidateOptionsResult.Fail(String.Join(Environment.NewLine,
+                    validationResults.Select(r => "DataAnnotation validation failed for members " +
+                        String.Join(",", r.MemberNames) +
```
Eilon: OK last nit, a space after the comma 😄

PR #274, opened 2018-09-04
Add OptionsValidator for eager validation

HaoK:
For https://github.com/aspnet/Home/issues/3386

Introduces the `IOptionValidator` which is used to 'eager' validate options. 

The default implementation uses a new `OptionsValidatorOptions` that just holds a list of actions.  A new method on `OptionsBuilder<TOptions>.ValidatorEnabled()` registers options for validation by the validator.

Example:
```C#
        [Fact]
        public void CanValidateMultipleOptionsEagerly()
        {
            services.AddOptions<ComplexOptions>("bool")
                .Configure(o => o.Boolean = false)
                .Validate(o => o.Boolean)
                .ValidatorEnabled();
            services.AddOptions<ComplexOptions>("int")
                .Configure(o => o.Integer = 11)
                .Validate(o => o.Integer != 11, "Not 11.")
                .ValidatorEnabled();
            services.AddOptions<ComplexOptions>("good")
                .ValidatorEnabled();
            services.AddOptions<ComplexOptions>()
                .ValidatorEnabled();

            var startupValidator = sp.GetRequiredService<IOptionsValidator>();

            var error = Assert.Throws<OptionsValidatorException>(() => startupValidator.Validate());
            var failures = error.ValidatorExceptions.ToArray();
            Assert.Equal(2, failures.Length);
            ValidateFailure<ComplexOptions>(failures[0], "bool", "A validation error has occured.");
            ValidateFailure<ComplexOptions>(failures[1], "int", "Not 11.");
        }
```

Note: this PR still leaves open the questions about whether we want to turn this on by default somehow, and make it easy to enable 'startup options validation', but that can be added in a follow on PR, since it would just be a matter of overloads that automatically call ValidatorEnabled, and also something in Hosting that would take care of invoking `IOptionsValidator.Validate` during startup.

Thoughts @ajcvickers @davidfowl @Eilon @DamianEdwards


Eilon:
Do we think we'd want a way for this to be enabled by default somehow? Who would want that and why? (I'm not doubting it, I just don't have a clear picture.)

HaoK:
Yeah so you could imagine that most people actually want their options validated on startup, so you get all your misconfigurations on app start instead of on a random request to some part of your app that uses some feature that is misconfigured.

I'm just not sure what the best way to expose that is yet.  We could add that as in hosting, some kind of OptionsValidatorStartupFilter / equivalent for generic hosts

ajcvickers (approved):

ajcvickers:
@HaoK @Eilon I feel like the Validator line in Configure is fine for now. We can always add sugar later based on feedback.

Eilon:
@ajcvickers you mean doing it in the template? Or telling customers they can do it on their own?

ajcvickers:
@Eilon I mean telling customers that this is what they should do if they want to a) use validation and b) validate eagerly.

Eilon:
That seems reasonable to me. Maybe worth doing a short review with Davian Fowlards?

HaoK:
Per review will move this code into the test project instead of shipping it for 2.2

HaoK:
@ajcvickers should I put this into 2.2 or just put it into master for 3.0 directly since this is just future looking test updates now

Eilon:
If it's useful to have as tests for 2.2, I'd go there first.

HaoK:
We already have a similar eager validation test in 2.2, it just uses a different implementation.

I actually think maybe we should leave this in PR form against master/3.0 as a prototype for comments instead of converting it to a test, its more visible as a potential future feature that way.  I can also include a reference to this PR as part of the documentation for options validation in 2.2

Eilon:
Totally up to you.

natemcmaster:
I'm closing this as the Options repo has merged into https://github.com/aspnet/Extensions. If this is something we still want, please re-open there.

PR #275, opened 2018-09-18
Search for errors instead of expecting strict match

HaoK:
Print out actual errors so we can see why this test fails in the future, for https://github.com/aspnet/Home/issues/3506

@Eilon @ajcvickers


HaoK:
Updated error validation to scan for matches to accommodate arbitrary order of data annotation errors

test/Microsoft.Extensions.Options.Test/OptionsBuilderTest.cs
```diff
@@ -387,15 +387,29 @@ public ValidateOptionsResult Validate(string name, ComplexOptions options)
...
+            if (errorsToMatch.Length == 0)
             {
-                errors = new string[] { "A validation error has occured." };
+                errorsToMatch = new string[] { "A validation error has occured." };
+            }
+            Assert.Equal(count, e.Failures.Count());
+            // Check for the error in any of the failures
+            foreach (var error in errorsToMatch)
+            {
+                var found = false;
```
Eilon: I bet this whole inner loop can be one LINQ statement. Might be easier to read. Up to you.

HaoK: Yeah good call, much more compact

